PR #17, opened 2020-12-15
[sdk] Add web preview support

IjzerenHein:
# Why

Adds support for web-preview (also known as the web-player) integration to the snack-sdk.

image

This PR adds auto-configuration and a transport for using the Snack web-player (v2). This code was previously implemented in snack-web and also in the example app. This consolidates the code in the snack-sdk and make it easy for anyone to use the web-player. This also provides a better upgrade path when we want to provide access to partners.

### Usage

```tsx
import * as React from 'react';
import { Snack } from 'snack-sdk';

export default () => {
  const webPreviewRef = React.useRef(null);
  const [snack] = React.useState(() =>
    new Snack({
      ...
      webPreviewRef,
    })
  );
  const { webPreviewURL } = snack.getState();
  return (
    <div>
      ...
      <iframe
        ref={(c) => (webPreviewRef.current = c?.contentWindow ?? null)}
        src={webPreviewURL}
        allow="geolocation; camera; microphone"
      />
    </div>
  );
};
```

# How

Adds a `webPreviewRef` init field, which when specified causes a `webplayer` transport to be created and the `webPreviewURL` field to receive the URL for setting the source on the iframe `<iframe src={webPreviewURL} />`.
The `webPreviewRef` value should be a react-like ref where `.current` refers to the contentWindow of the iframe. This design favours React but can also be used in non-react like environments.

- Only supported on SDK 40 and up (`webPreviewURL` will be undefined on earlier SDKs)
- Update example app (main source of docs)
- Update `./docs`
- Add webplayer tests

# Test Plan

- All tests succeed

### Manual test using example app

- Run `yarn build` in the snack-sdk
- cd `./example`
- `yarn reload`
- Open `http://localhost:8080`

### Manual test using Snack web-app

- Published to npm on `next`, version `3.3.0-rc.2`
- See https://github.com/expo/universe/pull/6639


byCedric (approved):
Native support for web-players looks awesome! Definitely a big improvement!

PR #69, opened 2021-02-03
[sdk] Add support for access token authentication

byCedric:
# Why

Adds basic support for `user: { accessToken: '...' }` to programmatically interact with stored Snacks.

# How

- Added helper for the `Expo-Session` or `Authorization: bearer ...` headers.
- Added tests for devsession to send authenticated requests

# Test Plan

- Create a new snack
- Add `snack.setUser({ accessToken: '...' })`
- Save the snack
- Snack should be saved on account by only using the access token


byCedric:
Clsoing in favor of #70

PR #70, opened 2021-02-03
[sdk] Add support for access token authentication

byCedric:
# Why

Adds basic support for `user: { accessToken: '...' }` to programmatically interact with stored Snacks.

# How

- Added helper for the `Expo-Session` or `Authorization: bearer ...` headers.
- Added tests for devsession to send authenticated requests

# Test Plan

- Create a new snack
- Add `snack.setUser({ accessToken: '...' })`
- Save the snack
- Snack should be saved on account by only using the access token

### DIFF
--- a/packages/snack-sdk/src/DevSession.ts
+++ b/packages/snack-sdk/src/DevSession.ts
@@ -1,6 +1,6 @@
 import { Logger } from './Logger';
 import { SnackState, SnackUser, SnackSendBeaconRequest } from './types';
-import { fetch } from './utils';
+import { createUserHeader, fetch } from './utils';
 
 export default class DevSession {
   private apiURL: string;
@@ -66,8 +66,8 @@ export default class DevSession {
       url,
       method: 'post',
       headers: {
-        ...(user?.sessionSecret ? { 'Expo-Session': user.sessionSecret } : {}),
         'Content-Type': 'application/json',
+        ...createUserHeader(user),
       },
     };
   }
--- a/packages/snack-sdk/src/utils.ts
+++ b/packages/snack-sdk/src/utils.ts
@@ -69,7 +69,7 @@ export function createUserHeader(user?: SnackUser): { [key: string]: string } {
   }
 
   if (user?.accessToken) {
-    return { Authorization: `bearer ${user.accessToken}` };
+    return { Authorization: `Bearer ${user.accessToken}` };
   }
 
   return {};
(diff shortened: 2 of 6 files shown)


packages/snack-sdk/src/DevSession.ts
```diff
@@ -66,8 +66,8 @@ export default class DevSession {
       url,
       method: 'post',
       headers: {
-        ...(user?.sessionSecret ? { 'Expo-Session': user.sessionSecret } : {}),
         'Content-Type': 'application/json',
+        ...createUserHeader(user),
```
IjzerenHein, line 70: Nice!

packages/snack-sdk/src/utils.ts
```diff
@@ -62,3 +62,15 @@ export function createError(config: {
...
   return error;
 }
+
+export function createUserHeader(user?: SnackUser): { [key: string]: string } {
+  if (user?.sessionSecret) {
+    return { 'Expo-Session': user.sessionSecret };
+  }
+
+  if (user?.accessToken) {
+    return { Authorization: `bearer ${user.accessToken}` };
```
IjzerenHein, line 72: Does it make any difference how `Bearer` is capitalized?

```suggestion
    return { Authorization: `Bearer ${user.accessToken}` };
```

IjzerenHein (changes requested):
Looks good 👍 

We should also add tests for Saving, which is probably the most important scenario to verify (much more than dev-sessions).

packages/snack-sdk/src/utils.ts
byCedric: You are right, it should be capitalized, afaik bearer tokens originates from OAuth. [In that spec](https://tools.ietf.org/html/rfc6750#section-2.1) it's defined with a capital. We also [use the capitalized one in CLI](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/ApiV2.ts#L205)

IjzerenHein:
friendly ping to @byCedric :)

byCedric:
@IjzerenHein done, added the same test case for `accessToken` and `sessionSecret` for snack saves.

IjzerenHein (approved):

IjzerenHein:
Awesome! Noticed that the tests weren't completing. This is because each snack that is `online`, also needs to be turned `offline`.

PR #90, opened 2021-02-19
improve Togglers visibility/redability, especially in Dark Mode

Simek:
# Why

After latest tweaks and changes switch Togglers (especially in Dark Mode) lose a bit o readability, and becomes a bit hard for user to distinguish at first glance between "On" and "Off" states in Dark Mode. 

The reason behind that is that, the "bullets" have used shared `selected` color, which in light theme is a primary color (vivid purple), but a white color in Dark Mode. 

# How

This PR ~introduced new `active`~ **reuses `primary`** color in the theme ~(which I hope can be utilized in few more places to improve readability in Dark Mode)~ to improve the Togglers visibility. In light mode, color will be the same as `selected`, but in Dark Mode it becomes less saturated, but still quite contrasting version of primary purple.

Additionally I have tweaked the Togglers background color, it's barely visible change in Light Mode, but in my opinion makes an significant improvement to the Togglers visibility in Dark Mode.

Since the PR have introduced the new color, the change should not affect any of existing elements or components.

In this PR I have also added `.idea` directory to the `.gitignore`, which contains the configuration files for my IDE.

# Test Plan

I have tested the changes on the `localhost`.

### Preview

Screenshot 2021-02-19 143012

Screenshot 2021-02-19 143028

## EDIT

After reading through and plaything more with the code I have discovered that exact the same color values, as newly introduced `active` color, provides just the first color in the list, which is `primary`. So I have reverted the addition, which makes this PR even simpler. :)


IjzerenHein (approved):
I like this improvement, thanks!

PR #91, opened 2021-02-19
[website] Unify keyboard shortcuts and appearance

Simek:
# Why

Currently, depending on the place, keyboard shortcuts could look a bit differently. Not all of them have the same color shade or were using same font-family. Also differentiating the appearance more between the label and shortcut could help some users better distinguish bot parts of item/menu.

Also I was thinking about, and testing, addition of "button frame" in two forms:
* regular keyboard button (border + small radius + double border/box-shad at the bottom)
* pill style button (border + 50% radius)

But I wasn't happy with any of this solutions. They have made the shortcuts eye-heavy and they draw too much of user attention, so I have completely opt-out adding any frame in this changeset (but the refactored code should make a bit easier future experiments).

# How

This PR aims to unify the appearance of all keyboard shortcuts across the Snack, aims to help improve HTML semantics by always wrapping shortcuts with `<kbd>` tag (hence `ShortcutLabel` refactor).

I have also tweaked a bit `EditorFooter` settings menu by removing the "Shortcuts" icon (it was a bit confusing in my opinion), adding separator and ensuring that whole "Shortcuts" item is clickable (like the items with the `Toggler` inside). 

Monospaced font have been used to differentiate even more shortcut from the menu/option label. The changes also applies to the Monaco editor context menu shortcuts.

# Test Plan

I have tested the changes on `localhost`.

### Demo

shkts

### DIFF
--- a/website/src/client/components/EditorFooter.tsx
+++ b/website/src/client/components/EditorFooter.tsx
@@ -136,18 +136,16 @@ export default function EditorFooter(props: Props) {
         label={<span className={css(styles.buttonLabel)}>Editor</span>}
         content={
           <>
-            <div className={css(styles.buttonItem, styles.buttonItemEditorPane)}>
-              <IconButton
-                small
-                title="Show keyboard shortcuts"
-                label="Shortcuts"
-                onClick={onShowShortcuts}>
-                <svg width="18px" height="18px" viewBox="0 0 18 18">
-                  <path d="M3,0 L15,0 C16.6568542,-3.04359188e-16 18,1.34314575 18,3 L18,15 C18,16.6568542 16.6568542,18 15,18 L3,18 C1.34314575,18 -2.46162913e-15,16.6568542 -2.66453526e-15,15 L-1.77635684e-15,3 C-1.97926296e-15,1.34314575 1.34314575,-5.83819232e-16 3,-8.8817842e-16 Z M7.41949521,8.19170984 L7.41949521,5 L6,5 L6,13 L7.41949521,13 L7.41949521,9.68911917 L7.71192341,9.44041451 L11.0992167,13 L13,13 L8.7232376,8.62176166 L12.975631,5 L11.0809399,5 L7.41949521,8.19170984 Z" />
-                </svg>
-              </IconButton>
-              <ShortcutLabel combo={Shortcuts.shortcuts.combo} />
+            <div
+              className={css(styles.buttonItem, styles.buttonItemEditorPane)}
+              onClick={onShowShortcuts}>
+              <IconButton small title="Show keyboard shortcuts" label="Shortcuts" />
+              <ShortcutLabel
+                combo={Shortcuts.shortcuts.combo}
+                className={css(styles.buttonItemShortcut)}
+              />
             </div>
+            <div className={css(styles.menuSeparator)} />
             <ToggleSwitch checked={fileTreeShown} onChange={onToggleFileTree} label="Files" />
             <ToggleSwitch checked={panelsShown} onChange={() => onTogglePanels()} label="Panel" />
             <ToggleSwitch checked={theme !== 'light'} onChange={onToggleTheme} label="Dark theme" />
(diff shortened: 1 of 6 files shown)


website/src/client/components/EditorFooter.tsx
```diff
@@ -136,18 +136,16 @@ export default function EditorFooter(props: Props) {
...
-              </IconButton>
-              <ShortcutLabel combo={Shortcuts.shortcuts.combo} />
+            <div
+              className={css(styles.buttonItem, styles.buttonItemEditorPane)}
+              onClick={onShowShortcuts}>
+              <IconButton small title="Show keyboard shortcuts" label="Shortcuts" />
+              <ShortcutLabel
+                combo={Shortcuts.shortcuts.combo}
+                className={css(styles.buttonItemShortcut)}
+              />
```
IjzerenHein, line 146: Shortcut label is displayed with a too small font here.

image

IjzerenHein (changes requested):
These a good improvements, thanks! 👍 

Readability of the shortcuts modal isn't great now. The shortcut key-codes are the most important thing in this screen and should have more emphasis. I suggest giving them a stronger appearance and even increasing the font-size here.

image

website/src/client/components/EditorFooter.tsx
Simek: At the first glance, it looks like big difference between macOS and Windows percentage font scaling. 

Thank you for catching those issues out, I will investigate the issue and potential improvement in near future :)

Simek:
@IjzerenHein From my research it looks like the `Fira` font family caused a size/scaling issue on macOS. 

I have decided to switch to the monospace fonts used for `code` tags in the Expo Docs - this will be only a temporary solution, because according to Jon comment (https://github.com/expo/styleguide/pull/4#issuecomment-783625697), big update for the `styleguide` will allow to share more styles between Expo project (hopefully fonts too 🙂).

To address the shortcuts visibility in the model I have introduced the `boxed` variant, which should make the key strokes more prominent. Additionally I have tweaked sizes a bit, to ensure that text is always readable and added few fixes for various minor style issues.

### Previews

#### Win
Screenshot 2021-02-23 121244

#### macOS

IjzerenHein (commented):
The boxed shortcuts look amazing, wow nice one! 😍

The shortcuts label in the popup still looks awfully small on my Mac though (running Chrome)

image

Simek:
> The shortcuts label in the popup still looks awfully small on my Mac though (running Chrome)
> 
> image

This is weird, can you try to reset the local cache and force reload the styles? I'm also using Chrome on my Mac, maybe there is a difference between fonts installed on our systems. Can you share the screenshot of "Rendered fonts" pane located at the bottom of Chrome Dev Tools -->Elements -> Computed view?

Simek:
@IjzerenHein From further testing and help from one more review it looks like the problem is with fonts still, it looks like `Monaco` is missing on your devices (despite Apple claims - https://developer.apple.com/fonts/system-fonts/ - I can also have this font on Big Sur) or macOS don't bother to load it and skips straight to `Menlo`,

Quotes from Wikipedia `Monaco` page:
> [...] Menlo was introduced as the default monospaced font instead of Monaco in Terminal and Xcode, However, Monaco remains a part of macOS [...]

For a moment, let's agree that's how it is, to improve the current situation I have added to most recent Apple monospaced font 'SF Mono', which (if loaded/present) should improve a bit the font size in menu. 

But to be honest, the proper fix for this situation will be to use of of the monospaced font from Google Fonts directory (like it has been done for the base font - https://github.com/expo/snack/blob/main/website/src/server/pages/Document.tsx#L64 - so the appearance would be consistent between systems) or even use Expo's custom monospace font which glimpse I can see in the Expo Docs (`expo-brand-mono`), but it's not loading on macOS or Win for me.

As I stated before, it would be also nice to have some `font-familiy` definition in Expo `styleguide` (but it might not solve the issue without embedding custom font, but definitely it would be a nice improvement).

CC: @jonsamp

IjzerenHein (approved):

IjzerenHein:
I've checked and on my system it's rendering these fonts

image


It looks like the special symbols are just way smaller than the regular characters:

image

So I don't think we can fix that, or try to.
Although it looks a bit small for that particular case on Mac, I think we should leave it as is. It looks pretty good and it's not worth putting in more time and effort.

Thanks for taking a look into this 👍

PR #93, opened 2021-02-20
dark mode: tweak hover color and add popup shadows

Simek:
# Why

Currently the Dark Mode hover color is quite contrasting, in comparison to the hover color in Light Mode. This also contributes to the lower readability of few UI elements while active or hovered (for example `EditorFooter` sections):

Screenshot 2021-02-20 183006

Screenshot 2021-02-20 182941

# How

In this PR I have changed the `hover` color value in `darkColors` to the shade a step darker, which improve a bit the readability and makes closer match to the Light Theme hover effect.

Screenshot 2021-02-20 183342

Additionally I have added the `popup` shadow back in the Dark Mode. Since shadows are based on pure black color (`0 2px 6px rgba(0,0,0,0.15)`) I see no backdrafts from turning them back on. The current value isn't ideal (shadows are not that prominent as in Light Mode), but they helps a bit distinguish UI elements in certain situations, few examples:

Screenshot 2021-02-20 184209

Screenshot 2021-02-20 192717

Maybe in Dark Mode popups should use `small` shadow, which is a bit more prominent (`0 5px 10px rgba(0,0,0,0.12)`), but it might be semantic not correct with goals behind `@expo/styleguide` package.

# Test Plan

I have tested the changes on the `localhost`.

### DIFF
--- a/website/src/client/components/ThemeProvider.tsx
+++ b/website/src/client/components/ThemeProvider.tsx
@@ -76,7 +76,7 @@ const darkColors: Colors = {
   'soft-text': colors.black,
   background: darkGray[900],
   content: darkGray[800],
-  hover: darkGray[600],
+  hover: darkGray[700],
   disabled: darkGray[600],
   selected: colors.white,
   'selected-text': darkGray[800],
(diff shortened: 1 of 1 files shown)


IjzerenHein (approved):
LGTM 😃

PR #94, opened 2021-02-20
"Edit details" popup button style tweaks

Simek:
# Why

Currently the "Edit details" button in the details popup it's hard to distinguish from  the regular text in Dark Mode. Additionally, despite being a clickable element, it doesn't contains the `hover` style. There is also a small issue in the `:active` state, where top border becomes `black` (at least on Windows Chrome).

# How

This PR tweaks a bit the appearance of "Edit details" button - switches the text color to `primary` (which affects only Dark Mode), add basic `hoover` color and rewrites the border style which seems to be the culprit for the border issue.

P.S. The `hover` effect readability in Dark Mode should improve after #93 merge.

# Test Plan

I have tested the changes on `localhost`.

# Preview
et


IjzerenHein (approved):
Thanks 😃

PR #95, opened 2021-02-21
QRCode: remove ability to select screen reader only URL

Simek:
# Why

Currently it's possible to use selection to make visible the Screen Reader only URL which is a part of `QRCode` component.

bug

# How

This PR includes small CSS fix (utilizing `user-select`) which fixes the issue, additionally I have extracted the existing inline styles from the component and replaced comment just by setting self-explanatory name to the style variable name.

# Test Plan

I have tested the changes on `localhost` using existing `dev` pipeline.


IjzerenHein (approved):
Awesome, thanks! 😃

PR #99, opened 2021-02-23
[website] fix Footer elements appearance while in error mode

Simek:
# Why

Currently, when error mode is activated red background is set to the `EditorFooter` and fonts colors are altered. This is causing some display issues when user hovers or activates a Footer button or SDK Switcher.

Screenshot 2021-02-23 at 14 16 43

# How

This PR adds few styles overwrites for states like `hover` and `active` to retain readability even in error mode. This probably can be improved in future utilizing the `hover` color with an alpha value (transparency).

Additionally in this PR I have fixed the `select` arrow character position and I have updated the `select` background so it's consistent with latest tweak for the `Togglers`.

Last small change fixes the `Toast` button colors in Dark Mode.

# Test Plan

I have tested the changes on localhost and have utilized Windows and macOS device to confirm that style changes are correct and don't cause an issues on one of the platforms.

### Windows

Screenshot 2021-02-23 141045

Screenshot 2021-02-23 141250

Screenshot 2021-02-23 141329

### macOS

Screenshot 2021-02-23 at 14 00 55

Screenshot 2021-02-23 at 14 00 49


IjzerenHein (changes requested):
Thanks 😃 

The toast en footer hover changes look great!

The SDK version switcher doesn't look right to me tbh. When the bottom-bar is blue or red, it stands out way too much. I think we should keep its background color the same as the bar, and only change the text-color when the bottom bar is blue or red.

Simek:
@IjzerenHein I have reverted SDK Switcher background color to transparent as you suggested and I have added a code to ensure the select border have the correct color in the error mode (previously, when Dark Mode was on, select have dark border, which do not match the overall error styling).

#### Preview
Screenshot 2021-02-25 230134
Screenshot 2021-02-25 230112

IjzerenHein (approved):
Looks great, thanks 😄

PR #110, opened 2021-02-26
[website] Increase infra to minimum of 4 pods

byCedric:
# Why

Discussed this in the response of [issue #1449](https://expo.pagerduty.com/incidents/PJJNR3W).

# How

- Updated `replicas` to `4`
- ~~Added minimum replicas (as per @quinlanj's suggestion)~~

# Test Plan

- Validate in GCP Kubernetes

### DIFF
--- a/website/deploy/production/deployment-replicas.yaml
+++ b/website/deploy/production/deployment-replicas.yaml
@@ -3,4 +3,5 @@ kind: Deployment
 metadata:
   name: snack
 spec:
-  replicas: 2
+  replicas: 4
+  minReplicas: 4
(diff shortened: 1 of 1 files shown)


nicknovitski (changes requested):
[Deployment resources](https://kubernetes.io/docs/reference/kubernetes-api/workloads-resources/deployment-v1/) don't have a `minReplicas` attribute.

[HPA resources](https://kubernetes.io/docs/reference/kubernetes-api/workloads-resources/horizontal-pod-autoscaler-v1/) do, and you could add one of those to the kustomization instead, while removing this file (no need to set the replicas number at all when an autoscaler will control it).

byCedric:
Thanks for the context @nicknovitski (cc @quinlanj )

IjzerenHein (commented):
Looks good to me, but I don't have enough knowledge about the topic in order to really tell :)

nicknovitski (approved):

PR #111, opened 2021-03-02
improve OS tabs appearance in Dark Mode, improve EmbeddedFooter

Simek:
# Why

Currently (like for switch) the OS tabs are using `selected` color, so in Dark mode it becomes white (in comparison to the purple in the Light Mode).

Screenshot 2021-03-02 182937

Also the OS tabs in the Embedded Snack are affected. Not only the color is not that readable, but the tabs are also clipped at the bottom (especially visible in the Dark Mode).

Screenshot 2021-03-02 182912

# How

This small PR changes the tabs background and border colors to `primary` which make the purple in the Dark Mode too.

Screenshot 2021-03-02 181941

Additionally I have fixed the clipping issues by removing the hardcoded height of `EmbeddedFooter`. This change results in even (`4px`) padding from all the sides, which improves the overall appearance and fixes the clipping.

Screenshot 2021-03-02 182153

# Test Plan

Snack run on the `localhost`.

### DIFF
--- a/website/src/client/components/Shell/EmbeddedFooterShell.tsx
+++ b/website/src/client/components/Shell/EmbeddedFooterShell.tsx
@@ -25,7 +25,6 @@ const styles = StyleSheet.create({
     transition: 'background .2s',
     padding: '0 4px',
     fontSize: 12,
-    height: 26,
   },
 
   footerLoading: {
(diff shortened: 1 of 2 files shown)


IjzerenHein (approved):
This looks good, thanks!

PR #113, opened 2021-03-08
[website] replace DeviceID hint screenshot with in-code example

Simek:
# Why

Currently a screenshot is used for an example of the the DeviceID location, which doesn't look that great in dark mode.

Also the screenshot seems to be a bit outdated, since there is also an additional information about SDK listed in Expo Go.

Screenshot 2021-03-08 204356

# How

This small PR replaces the used image asset with an in-code example, to improve the appearance in dark mode. This should also allow for easier changes in the future, if such will be needed.

### Preview
Screenshot 2021-03-08 204416
Screenshot 2021-03-08 204040

# Test Plan

I have developed and tested the changes using Snack on localhost.


IjzerenHein (approved):
Thanks again @Simek, this is a great improvement!

PR #122, opened 2021-04-19
[website] Fix logs panel text wrapping

Simek:
# Why

Currently logging long value, like arrays or objects keys results in vertical scroll in the Logs panel and shifted a bit content.

### Before

Screenshot 2021-04-19 133749

# How

This PR fixes the issues by adding few CSS lines. 

~Unfortunately `aphrodite` [package types](https://github.com/Khan/aphrodite/blob/master/typings/css-properties.d.ts#L1740) do not list `whiteSpace: 'break-spaces` option so I had to add the line ignore. `break-spaces` is not that new property and it's [widely supported](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space), but the `aphrodite` seems to be no longer maintained - types definitions are 3+ years old.~

After testing a bit, it looks like that `whiteSpace: 'normal'` yield the same layout result as `'break-spaces'`, so I have switched the values to avoid TS ignore comment. 🙂 

### After

Screenshot 2021-04-19 133643

# Test Plan

Running Snack website on `localhost`.

### DIFF
--- a/website/src/client/components/EditorPanelLogs.tsx
+++ b/website/src/client/components/EditorPanelLogs.tsx
@@ -94,6 +94,7 @@ const styles = StyleSheet.create({
     display: 'flex',
     flexDirection: 'row',
     flexWrap: 'wrap',
+    whiteSpace: 'pre',
   },
   item: {
     marginRight: 8,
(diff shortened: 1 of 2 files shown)


IjzerenHein (approved):
Awesome =D

PR #121, opened 2021-04-16
[snack-sdk] Upgrade to Expo SDK 41

byCedric:
# Why

Upgrade to support Expo SDK 41

# How

Followed steps [described here](https://github.com/expo/snack/pull/109/files#diff-5172fce9dfcb48632737b8249989e0d298304200575759ea04ced389284da81eR41).

# Test Plan

Staging should be updated with the changes required for SDK 41.

- Test snack.expo.io (with no id!)
  - Open unsaved snack in app, make some changes, then hit save
- Test existing snack.expo.io/ID (https://snack.expo.io/r1Xun_7eb)
  - Open that in the app, make some changes, save, and reload
- Test existing permanent snack snack.expo.io/@username/project
  - Open that in the app, make some changes, save, and reload
- Open snack directly on phone without website open (https://expo.io/@snack/r1Xun_7eb)
- Open permanent snack directly on phone without website open (https://expo.io/@username/project)
- Open `website/snack-embed.html` to test embeds

PR #124, opened 2021-04-19
[snack-sdk] Release Expo SDK 41

byCedric:
# Why

This is a stacked PR on top of #123, but for the `snack-sdk` instead. 

# How

- Dropped support for SDK 37
- Switched to SDK 41 by default
- Updated all tests and snapshots to include 38 or higher

# Test Plan

- `$ yarn test`
- Make sure staging is working as expected using the test plan from #121

### DIFF
--- a/packages/snack-sdk/src/__tests__/__snapshots__/dependencies-test.ts.snap
+++ b/packages/snack-sdk/src/__tests__/__snapshots__/dependencies-test.ts.snap
@@ -137,8 +137,8 @@ Object {
 exports[`dependencies resolves * after enabling 1`] = `
 Object {
   "expo-constants": Object {
-    "version": "~9.0.0",
-    "wantedVersion": "~9.0.0",
+    "version": "~9.1.1",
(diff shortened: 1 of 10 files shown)


packages/snack-sdk/src/__tests__/__snapshots__/dependencies-test.ts.snap
```diff
@@ -200,32 +200,35 @@ Object {
...
+    "wantedVersion": "~8.2.1",
   },
 }
 `;
 
 exports[`dependencies updates preloaded module version when changing SDK version 2`] = `
 Object {
   "expo-av": Object {
-    "handle": "snackager-1/expo-av@8.2.1",
+    "error": [Error: Failed to resolve dependency 'expo-av@~8.6.0' (Package 'expo-av@~8.6.0' not found in the registry)],
```
byCedric, line 225: This shouldn't be here, let me fix this asap!

brentvatne (approved):

PR #125, opened 2021-04-20
[website] set default SDK to v41

Simek:
Refs #123

# Why

SDK v41 has been added to Snack, but v40 is still set as a default version on Snack load.

# How

This PR updates the `DEFAULT_SDK_VERSION` to v41, which is responsible for default the selected version on the website.

# Test Plan

Change has been tested by running Snack website on `localhost`

# Preview

Screenshot 2021-04-20 134138

CC @byCedric


Simek:
Oh, I didn't see the commits or edited description in v41 release PR, so this change looks like invalid (at least for now).

PR #128, opened 2021-04-22
[snackager] Add package specific cache-busting

IjzerenHein:
# Why

The `react-native-reanimated2` native bundles are invalid because they internalized `react-native/Libraries/Renderer/shims/ReactNative`. This was fixed by PR #126 This PR adds support for package specific cache-busting to Snackager and invalidates the current caches for `react-native-reanimated2`.

# How

- Extend the cache-handle to include an additional suffix (e.g. `-1`) to invalidate current caches
- Bust cache for `react-native-reanimated`

# Test Plan

- Verified locally
- `yarn start`
- Open website and add `react-native-reanimated` to the dependencies
- Watched how Snackager rebuild the package
- Removing and re-adding the package returned the newly cached package

### DIFF
--- a/snackager/src/cache-busting.ts
+++ b/snackager/src/cache-busting.ts
@@ -1,0 +1,21 @@
+const cacheBusting: { version: number; packages: { [name: string]: number } } = {
+  /**
+   * Main version. Incrementing this causes all cached bundles
+   * to be invalidated and rebuild when requested.
+   * */
+  version: 1,
+  /**
+   * Package specific invalidations.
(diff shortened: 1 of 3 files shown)


byCedric (approved):
Would be nice to have some documentation about this too, I can help with that 😄

IjzerenHein:
> Would be nice to have some documentation about this too, I can help with that 😄

Yeah that's a good idea. I tried naming the file so ppl would understand and added some comments already.
A top level FAQ for Snackager would be very useful and allow us to offload more the OS community

PR #133, opened 2021-04-29
Use staging environment variable in runtime workflow

byCedric:
# Why

This enables the GH action to authenticate with staging, instead of production.

# How

- Added global environment variable for the job

# Test Plan

- See if CI passes


IjzerenHein:
Merged with #132

PR #139, opened 2021-05-12
[k8s] deploy snack on expo.dev

FiberJW:
# Why

We want to deploy Snack on snack.expo.dev so that it can share cookies with users logged in on expo.dev

# How

- added new host entries to staging and production ingress config.
- [ ] deploy to staging to test if these changes are working appropriately (I'm not sure of how to do this, guidance would be really appreciated)

# Test Plan

After this change, snack.expo.dev should work and render the snack UI

### DIFF
--- a/website/deploy/production/ingress-spec.yaml
+++ b/website/deploy/production/ingress-spec.yaml
@@ -6,9 +6,17 @@ spec:
   tls:
   - hosts:
     - snack.expo.io
+    - snack.expo.dev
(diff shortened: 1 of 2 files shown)


nicknovitski:
This change will get deployed to staging by skaffold after being merged, and since there's already a dns record for `*.expo.dev`, it should start routing traffic right away.

nicknovitski (approved):

PR #142, opened 2021-05-20
[website] Update appetize preview to iOS 14, iPhone 12 and Pixel 4

IjzerenHein:
# Why

Appetize preview uses iOS 13 and older iPhone/Android models. This PR updates the Appetize previews to use newer phone models; and to iOS 14. These new modals (iPhone12 & Pixel4) showed rounded corners in embedded mode, which looked off. For that reason, the bezels were enabled in embedded (see screenshots at the bottom). As these newer modals have a larger screen surface, embedded mode now requires a larger height to display the full preview. When the embedded iframe is too small, the bottom part of the preview is cutoff (as was also done before).

# How

- Update appetize device preview to use iPhone12 and Pixel4
- Show bezels on embedded mode
- Updated UI layout to accommodate the new device models
- Tests were updated

# Test Plan

## non embedded

image

image

## embedded

image

image

### embedded (limited height)

image

image

### bezels were added to embedded as new models showed rounded corners which looked weird

image

image


byCedric (approved):
Looks good on my machine too

PR #144, opened 2021-05-20
[ci] Runtime deployment github actions

IjzerenHein:
# Work in progress

# Why

This PR adds automatic deployment of the Snack runtime to the staging and production environment (web-player & native runtime). Previously this needed to be done by hand, which was cumbersome and time consuming.

# How

- Auto deploy to staging upon pushing to main
- Deploy to production by running the `deploy` workflow in `RuntimeProduction`
- Native runtime deploy has been verified
- Web-player deploy is Work in progress, requires S3 credentials config to work

# Test Plan

- [X] Verified deployment of the native runtime to staging
- [X] Verified slack notification on staging
- [ ] Verified web-player deployment to staging (FAIL: requires S3 credentials) 
- [ ] Verified deployment of the native runtime to production (should work, but not yet verified)
- [ ] Verified slack notification on staging (should work, but not yet verified)
- [ ] Verified web-player deployment to staging (FAIL: requires S3 credentials)

### DIFF
--- a/.github/workflows/runtime-production.yml
+++ b/.github/workflows/runtime-production.yml
@@ -36,6 +36,7 @@ jobs:
       - uses: expo/expo-github-action@v6
         with:
           expo-version: 4.x
+          expo-cache: true
           token: ${{ secrets.EXPO_PRODUCTION }}
 
       - run: yarn install --frozen-lockfile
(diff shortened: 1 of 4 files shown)


byCedric:
Blocked by an issue in our Terraform file for this. Still need to finish other stuff before I can continue on that.

PR #147, opened 2021-05-26
[website] Remove file import JS limitation

byCedric:
# Why

Right now, users can only upload JS files through the "Import files" modal. I don't think we should limit it to JS only, as users should be able to upload images and maybe JSON/txt files too.

# How

Removed the `accept="js"` from the file import modal.

# Test Plan

- See if we can upload more types of files after this change, e.g. assets and code.


IjzerenHein (approved):

IjzerenHein:
Verified locally and works great. Yeah that limitation didn't make any sense 👍

PR #150, opened 2021-05-27
[website] Replace snack-search with API docs search

IjzerenHein:
# Why

Searching for snacks is in almost all cases unhelpful as there are many unfinished/draft/experiment and unnamed snacks. It was determined that this feature is not helpful and that searching for APIs that you can use is more useful and can be achieved will relatively low effort. This PR therefore removes the Snack search button and modal and replaces it with a search-bar for searching the API docs. The search uses the Expo Algolia docs-index. When clicking on a result, the search-result is opened in a new tab. Searches are also SDK specific and will show results for the selected SDK only.

image

image

image

image

# How

- Remove Snack search
- Add API-search search-bar 
- Search is themed and supports dark-mode
- Re-initialize search-bar when changing SDK
- Searches only versioned resources and not any getting started walkthroughs or tutorials
- Search-bar is only shown in full-mode, not in embedded mode

# Test Plan

- All tests succeed
- Verified that clicking a result opens the correct SDK page
- Verified that changing the SDK version returns search results for that SDK and links are correct
- Dark mode is correctly supported

image


brentvatne (approved):
rip broken feature

website/src/client/components/Search/SearchBar.tsx
```diff
@@ -0,0 +1,126 @@
...
+        facetFilters: [[`version:v${this.props.sdkVersion}`]],
+      },
+      handleSelected: (
+        input: any,
+        _event: any,
+        suggestion: any,
+        _datasetNumber: any,
+        context: any
+      ) => {
+        // Prevents the default behavior on click and rather opens the suggestion
```
brentvatne, line 48: :+1:

PR #154, opened 2021-06-01
feat(import-repo-modal): note private repos cause errors

schwartzadev:
# Why
I attempted importing a private Github repo and got an error message that was not related to the repo's visibility.

# How
Updated the error copy

# Test Plan

Repro:
1. Make an expo project
2. Place that expo project in a private github repo
3. Attempt to import that private github repo

We should probably aim to have more verbose errors in the future, but I think this is an improvement in the near-term.


IjzerenHein:
Hi! Thank for the PR, this is a good suggestion. This has been fixes as part of PR #157.
Cheers

PR #155, opened 2021-06-07
[website] redirect to new domain (snack.expo.dev) if flag is enabled

FiberJW:
# Why

To handle the transition, I want to introduce a flag that we can turn on/off to run redirect code.

# How

- adds `WEBSITE_URL` and `SNACK_DOT_DEV_SERVER_URL` environment variables for determining origin and destination redirect paths
- adds `isExpoDevEnabled()` which enables the expo.dev redirect flow for development and staging
- adds `transitionToExpoDev/undoTransitionToExpoDev` which redirects snack.expo.io to snack.expo.dev and the inverse.

# Test Plan

- Go to `~/.chalet/servers` and duplicate the `snack.expo` file as `snack.expo.io`
- Run snack website
- Visit snack.expo.io.test and be redirected to snack.expo.io


website/src/server/routes.tsx
```diff
@@ -32,6 +33,10 @@ const createChannel = customAlphabet(
 );
 
 const render = async (ctx: Context) => {
+  const transitionIsSuccessful = transitionToExpoDev(ctx);
+
+  if (transitionIsSuccessful) return;
+
```
IjzerenHein, line 39:
```suggestion
if (redirectToDevDomain(ctx)) {
  return;
}
```

IjzerenHein (changes requested):

website/src/expo-dev-migration/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    isExpoDevRedirectEnabled &&
+    `${ctx.protocol}://${ctx.hostname}` === process.env.SNACK_DOT_DEV_SERVER_URL &&
+    process.env.SNACK_SERVER_URL
+  ) {
+    ctx.redirect(`${process.env.SNACK_SERVER_URL}${ctx.req.url}`);
+    return true;
+  } else {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
IjzerenHein: Okay gotcha, that's sound reasoning and good preparation. Considering the temporary nature, I would add a short comment that this function will be removed after the transition. It's good practise and helps as a reminder to yourself and others in case something comes up and this is idle for a while.

IjzerenHein (changes requested):
Looks great. The snack website has some duplicates unfortunately, and the new env-vars also need to be added to `skaffold.yml`😅

website/skaffold.yaml
```diff
@@ -18,6 +18,7 @@ build:
           APP_VERSION: staging-{{.GITHUB_SHA}}
           DEPLOY_ENVIRONMENT: staging
           IMPORT_SERVER_URL: https://staging.snackager.expo.io
+          LEGACY_SNACK_SERVER_URL: https://staging.snack.expo.dev
```
IjzerenHein, line 21: In order to test the redirects on staging, I think we should set this to `staging.snack.expo.io`

IjzerenHein (changes requested):

website/deploy/staging/snack.env
```diff
@@ -1,6 +1,7 @@
 SERVER_URL=https://staging.expo.dev
 API_SERVER_URL=https://staging.exp.host
 SENTRY_ENVIRONMENT=staging
+LEGACY_SNACK_SERVER_URL=https://staging.snack.expo.dev
```
IjzerenHein, line 4: Minor issue, this is not even used on the pods I think and I'm not sure even locally, but lets fix it

```suggestion
LEGACY_SNACK_SERVER_URL=https://staging.snack.expo.io
```

IjzerenHein (approved):

website/deploy/staging/snack.env
IjzerenHein: Feel free to merge 👍

PR #156, opened 2021-06-07
[website]  use *.expo.dev urls when transition is enabled

FiberJW:
# Why

The links/endpoints we reference should also change from expo.io to expo.dev when we start the redirect.

# How

- added `getWebsiteURL()` and `getSnackWebsiteURL()` which return *.expo.dev for development and staging and *.expo.io
- replaced `process.env` URL references with the getter function (other than the transition file, which relies on the constant values)

# Test Plan

The links and endpoints we hit should point to expo.dev and work as expected in dev/staging.


IjzerenHein (changes requested):

IjzerenHein:
I believe the build is failing because the new env-var are missing from `skaffold.yml`

website/deploy/staging/snack.env
```diff
@@ -1,3 +1,4 @@
+LEGACY_SERVER_URL=https://staging.expo.dev
```
IjzerenHein, line 1:
```suggestion
LEGACY_SERVER_URL=https://staging.expo.io
```

website/skaffold.yaml
```diff
@@ -13,6 +13,7 @@ build:
         dockerfile: website/Dockerfile
         buildArgs:
           node_version: 12.18.4
+          LEGACY_SERVER_URL: https://staging.expo.dev
```
IjzerenHein, line 16:
```suggestion
          LEGACY_SERVER_URL: https://staging.expo.io
```

IjzerenHein (changes requested):

IjzerenHein (approved):
LGTM :)

PR #158, opened 2021-06-10
[runtime] Remove installationId

IjzerenHein:
# Why

Removes the deprecated `Constants.installationId` and replaces it with a persisted randomly generated Id.

# How

- Use expo-random to generate a one time random id (hex 32 chars) and store in async/local-storage
- Remove `installationId` usage and warning

# Test Plan

- Loaded Snack runtime into Expo Go (iOS) and logged device-id to console
- Confirmed that the device-id remains the same after reloading the Snack runtime (by shaking and reloading)
- Loaded Snack web-player into the Snack website (using the localhost option)
- Confirmed that the device-id remains the same when refreshing the web-player

### DIFF
--- a/runtime/src/Analytics.tsx
+++ b/runtime/src/Analytics.tsx
@@ -10,10 +10,16 @@ Amplitude.initializeAsync(AMPLITUDE_KEY);
 
 let loggedReceivedFirstCode = false;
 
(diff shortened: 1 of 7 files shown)


byCedric (approved):
Noice

PR #160, opened 2021-06-11
[runtime] Allow web-player on sizze domain

IjzerenHein:
# Why

To allow running the Snack web-player on the sizze.io domain.

@Tozare

# How

- Add `https://dashboard.sizze.io` to list of allowed domains

# Test Plan

- Unable to test, needs to be performed by sizze.io after deployment

### DIFF
--- a/runtime/src/Messaging.web.tsx
+++ b/runtime/src/Messaging.web.tsx
@@ -48,6 +48,8 @@ const allowedOrigins = [
   'https://jackdaw.codecademy.com',
   'https://staging.codecademy.com',
   'https://production.codecademy.com',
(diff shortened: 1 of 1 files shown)


byCedric (approved):
We also need to whitelist our API, this will only allow some communications but storing wont work I think.

IjzerenHein:
Yeah that's right, saving will not yet work. We can add it in case they need it

Tozare:
Dear @IjzerenHein @byCedric ,
We are developing second version of Sizze on new domain. So, can you, please, add app.sizze.io to list of allowed domains too?

PR #163, opened 2021-06-21
[snackager] Fix react-dom not externalized in web-bundles

IjzerenHein:
# Why

Fixes web-bundles that use `react-dom` directly such as `@material-ui` (and indirectly `@drafbit/ui`). In these cases `react-dom` was falsly included **inside** the bundle which conflicted with `react-dom` which was already included in the Snack runtime using `react-native-web`. This PR resolves the problem in Snackager by ensuring that `react-dom` is always externalized.

**After the fix**

image

# How

- Always externalize `react-dom`
- Invalidate cache for `@draftbit/ui`
- Update tests

# Test Plan

- Bump cache-bust number for `@drafbit/ui`
- `yarn start`
- Open `http://snack.expo.test/` and import Snack from production: https://snack.expo.io/@peterpme/(draftbit)-onboarding---kid-info
- Load web-preview and click on date (see screenshot in #why)
- Run `yarn test` to verify all tests succeed

### DIFF
--- a/snackager/src/__integration-tests__/__snapshots__/bundler.test.ts.snap
+++ b/snackager/src/__integration-tests__/__snapshots__/bundler.test.ts.snap
@@ -420,7 +420,7 @@ Object {
           "react-native/Libraries/BatchedBridge/BatchedBridge",
           "react-native-svg",
         ],
-        "size": 294732,
+        "size": 294734,
       },
     },
     "ios": Object {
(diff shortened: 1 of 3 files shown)


byCedric (approved):
LGTM! The only worry I have is that we have to clear the cache for almost every library. If any of the prebuilt packages contains `react-dom`, they are invalid now or in the future once we upgrade react right?

IjzerenHein:
Only libraries that explicitly `react-dom` would be affected. As this is the first time I've seen this problem, it might not be as wide-spread as you think.

PR #164, opened 2021-06-23
[snackager] Add basic support for mjs packages

byCedric:
# Why

We are having issues with `aws-amplify` at the moment, it never gets build because of issues related to mjs files:

```
Can't reexport the named export 'isIntrospectionType' from non EcmaScript module (only default export is available)
```

# How

This issue is related to https://github.com/graphql/graphql-js/issues/1272. 

- ~~Applied the fix from that issue and it seems to work nicely.~~ That didn't work, probably because it relies on the node env, and node 12 isn't exactly the "latest" node.
- Let babel handle the mjs files instead, this seems to work for node 12 too.

# Test Plan

- `$ yarn bundle graphql@14.0.0` -> should fail before the fix
- `$ yarn bundle graphql@14.0.0` -> should succeed after the fix


snackager/src/__integration-tests__/__snapshots__/bundler.test.ts.snap
```diff
@@ -200,6 +200,61 @@ Object {
...
+          "./validation",
+          "./error",
+          "./utilities",
+        ],
+        "size": 20120,
+      },
+    },
+    "web": Object {
+      "bundle.js": Object {
+        "externals": Array [
```
byCedric, line 238: wait, I don't think these externals are right. Shouldn't they be built into the package?

byCedric:
Ok so the issue here (I think) is in the `mainFields`. When I pull my changes related to `mjs` support, _and_ remove the `modules` from the main fields, it compiles `graphql@14`. When I only add the `modules` back without mjs change, it fails.

We tell webpack to use the `modules` without bundling it properly. I'm not sure of we should just remove `modules` from the `mainFields` or if we should add the `mjs` with strange externals to snackager. @IjzerenHein

IjzerenHein:
@byCedric Yeah those externals are definitely not good. The outputted bundle is also too small and doesn't contain any relevant code. I've looked into this and have tried upgrading to webpack@5 to see whether that fixes anything, but it didn't.
I fiddled a while and finally found a solution though. This worked for me:

```
{
          test: /\.mjs$/,
          resolve: {
            fullySpecified: false,
          },
        },
```

You'll also need to remove the `m?` modification from the babel-loader.

I noticed that `include: /node_modules/` caused a problem not being able to find the package files itself. And `type: 'javascript/auto',` didn't seem to do much. The `fullySpecified: false` seems to fix an issue with `index.mjs` that can otherwise not be found.

byCedric:
@IjzerenHein is that with Webpack 5 by any chance? Webpack really don't like that resolve property, it just blocks on invalid config after forcing the property.

Screenshot 2021-06-25 at 14 37 43

IjzerenHein:
@byCedric Oh dang really?! Yeah I tested that on the webpack@5 branch, so must be a webpack 5 feature 😅

byCedric:
I'm afraid it is, https://github.com/webpack/webpack.js.org/issues/3975

IjzerenHein:
Hmm. Might be worth trying to switch to the webpack@5 branch. All packages except for `@drafbit/ui` seem to be building fine, although I haven't actually tested the produces bundles. They are all similar in size (slightly smaller) and have all the same externals, with a good markers of working bundles. `@drafbit/ui` has several unexpected externals with the `web` bundle which I haven't been able to figure out why

byCedric:
I think I kind-of mimicked the `fullySpecified: false` property by using my earlier fix, with a `.mjs` file extension added to the resolver. I'll push it and see what you think 😄

byCedric:
I'll take a look at webpack 5 to see if I can find anything why Draftbit doesn't build. But that would probably be somewhere Monday or Tuesday.

IjzerenHein (approved):
I've verified locally and works great :D
LGTM

PR #165, opened 2021-06-24
 [website][debug] add debug code to figure out why .dev transition isn't working in staging

FiberJW:
# Why

The snack.expo.io -> snack.expo.dev redirect isn't working as expected in staging, so I want to learn more about why since the code implementation seems simple.

# How

Added logging around env vars and execution flow (or the lack of execution).

## Test Plan

After this is deployed and merged in staging, I should be able to read the logs with `ks logs <snack-pod>` and/or `ks attach <snack-pod> -i`


IjzerenHein (approved):
Looks good, will get the job done. Might be useful to prefix all your log statements with a string that you can easily and uniquely search for using the Google console.

PR #169, opened 2021-06-25
[website] remove .dev redirect debug statements

FiberJW:
# Why

With the help of logging, I figured out that `app.proxy = true` was the key to getting the redirect code working, so the debug statements I added have now served their purpose and now should be removed.

# How

Removed all `console.log('.DEV Redirect...` statements.


IjzerenHein (approved):
👍

PR #170, opened 2021-07-01
[snackager] Use secure cloudfront url for snackager

byCedric:
# Why

Font resources from the `expo-google-fonts` package aren't loaded atm. This is due to mixed resources (HTTP vs HTTPS). This is a test to enable it on staging only.

# How

Updated the k8s staging envvars.

# Test Plan

See if this [snack loads properly](https://snack.expo.io/@peterpme/rebellious-orange) on staging.

### DIFF
--- a/snackager/k8s/staging/snackager.env
+++ b/snackager/k8s/staging/snackager.env
@@ -1,4 +1,4 @@
-CLOUDFRONT_URL=http://ductmb1crhe2d.cloudfront.net
+CLOUDFRONT_URL=https://ductmb1crhe2d.cloudfront.net
 IMPORTS_S3_BUCKET=snack-git-imports-staging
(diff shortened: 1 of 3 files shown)


IjzerenHein (approved):
In order to fix this for `@expo-google-fonts`, we should probably invalidate the cache for `@expo-google-fonts`, otherwise Snackager just returns the already generated bundles. We might need to introduce some kind of wildcard syntax to invalidate all `@expo-google-fonts/*` packages

IjzerenHein:
@byCedric I've added cache-busting for `@expo-google-fonts\*`. Can you review this?

byCedric (commented):
LGTM, can't approve it since it's my own PR.

IjzerenHein (approved):

PR #175, opened 2021-07-02
[sdk] Drop support for SDK 38

byCedric:
# Why

With the arrival of SDK 42, we say goodbye to SDK 38.

# How

- Followed the upgrade guide in https://github.com/expo/snack/blob/main/docs/expo-sdk-upgrade.md and moved everything from SDK 38.
- Kept the default SDK version at `41.0.0` so we can have one more pass if SDK 42 works as expected.

# Test Plan

For `snack-sdk` and `website`:

- `$ yarn lint`
- `$ yarn test`

Manual test to see if Snack works with `yarn start` in root.


byCedric:
Only tasks left after this PR:
- Release a new `snack-sdk` version, including SDK 42 and excluding SDK 38
- Set the default SDK to 42 instead of 41.

website/src/client/configs/sdk.tsx
```diff
@@ -4,12 +4,12 @@
  * This is useful when deploying the website  with preliminary support for a new SDK version.
  */
 export const versions: { [version: string]: boolean } = {
-  '38.0.0': true,
+  '38.0.0': false, // dropped with sdk 42.0.0
```
IjzerenHein, line 7: I think this will fail as the types from website and snack-sdk are now different. Should just remove this as SDK 38 won't work anymore after this PR

IjzerenHein (approved):
Looks good and thanks for updating all the tests!
Only 1 comment. I expected that to fail the build but it didn't as the types in website are probably a superset of the types in snack-sdk. We should probably make that more tightly coupled in a future PR

website/src/client/configs/sdk.tsx
byCedric: Removed it and used the `SDKVersion` type in the object!

PR #176, opened 2021-07-08
[sdk] Set default SDK to 42 and update changelog

IjzerenHein:
# Why

Sets the default Expo SDK for snack-sdk to 42.0.0, as well as updates the changelog and makes some type improvements to website. Once completed, the snack-sdk will be published to NPM with these changes.

# How

- Set default SDK in snack-sdk to 42.0.0
- Set default SDK in website to 42.0.0
- Update snack-sdk changelog
- Stricter types for SDK_VERSION constants in website
- Update snackager test snapshots

# Test Plan

- snack-sdk `yarn lint && yarn tsc`
- snack-sdk `yarn test`
- website `yarn lint && yarn tsc`
- website `yarn test`
- snackager `yarn test`

### DIFF
--- a/website/src/client/configs/sdk.tsx
+++ b/website/src/client/configs/sdk.tsx
@@ -12,5 +12,5 @@ export const versions: Record<SDKVersion, boolean> = {
   '42.0.0': true,
 };
 
-export const DEFAULT_SDK_VERSION = '41.0.0';
-export const TEST_SDK_VERSION = '39.0.0';
+export const DEFAULT_SDK_VERSION: SDKVersion = '42.0.0';
+export const TEST_SDK_VERSION: SDKVersion = '39.0.0';
--- a/packages/snack-sdk/CHANGELOG.md
+++ b/packages/snack-sdk/CHANGELOG.md
@@ -4,7 +4,12 @@
 
 ### 🛠 Breaking changes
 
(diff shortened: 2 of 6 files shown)


website/src/client/configs/sdk.tsx
```diff
@@ -12,5 +12,5 @@ export const versions: Record<SDKVersion, boolean> = {
   '42.0.0': true,
 };
 
-export const DEFAULT_SDK_VERSION = '41.0.0';
-export const TEST_SDK_VERSION = '39.0.0';
+export const DEFAULT_SDK_VERSION: SDKVersion = '42.0.0';
```
IjzerenHein, line 15: snack-sdk also defines a "default" SDK-version. This const has always confused me a bit and feels duplicate. In my opinion we can just remove this and use the default SDK version from `snack-sdk`. @byCedric what do you think?

byCedric: The only scenario where this could go wrong is when we release an `x.x.x-rc.0` version. That shouldn't be set on the website instantly right? When we do that, we have to be able to use a different default version on the website

IjzerenHein: Yeah I think you're right. Lets keep it as is 👍

byCedric (approved):

PR #178, opened 2021-07-08
[website] Run npx browserslist@latest --update-db

IjzerenHein:
# Why

Fixes warning that `caniuse-lite` is outdated.

# How

- Run `npx browserslist@latest --update-db` in root

# Test Plan

- All tests should succeed
-


byCedric:
Not sure what's happening in CI, but this change seems good to me. Maybe some snapshots might fail as well (if it includes React components and styling)

byCedric (approved):

PR #182, opened 2021-07-09
[ci] Run runtime CI workflows for sdk-* branches

IjzerenHein:
# Why

This PR updates the runtime CI actions to allow deployment to staging & production from `sdk-*` branches. The changes for this PR were tested on the `sdk-41` branch through #180 and #181. This PR ports them to the main branch so the changes will apply for future sdk versions as well.

# How

- Run generic Runtime workflow on any pull-request (instead of only PRs for `main`)
- Updated staging workflow to deploy when pushing to main or sdk-* branches
- Updated production workflow to deploy when pushing to main or sdk-* branches

# Test Plan

- Tested through PR #180 and #181
- Action that confirms the generic `Runtime` action runs on branches other than `main` 
  -https://github.com/expo/snack/actions/runs/1014691626
- Action that confirms succesful deployment of Runtime 41 to staging:
  - https://github.com/expo/snack/actions/runs/1014708028
- Actions that confirm succesful deployment of Runtime 41 to production:
  - https://github.com/expo/snack/actions/runs/1014748387
  - https://github.com/expo/snack/actions/runs/1014765302

### DIFF
--- a/.github/workflows/runtime-staging.yml
+++ b/.github/workflows/runtime-staging.yml
@@ -6,7 +6,9 @@ defaults:
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'sdk-**'
     paths:
       - .github/workflows/runtime-staging.yml
       - runtime/**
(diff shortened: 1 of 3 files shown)


.github/workflows/runtime-staging.yml
```diff
@@ -42,23 +44,21 @@ jobs:
...
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_RUNTIME_KEY_STAGING }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_RUNTIME_SECRET_STAGING }}
 
       - name: Deploy native runtime
-        if: github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
         run: yarn deploy:staging
 
       - name: Notify Slack
         uses: 8398a7/action-slack@v3
-        if: always() && github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
```
byCedric, line 57: If you want to make the "Only from `main` or `sdk-` branches" foolproof, [we could use something like `startsWith(github.ref, 'refs/heads/sdk-')`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#startswith)

byCedric (approved):

.github/workflows/runtime-staging.yml
IjzerenHein: I think that's a good idea. Especially if we want to merge some of the workflows later on.
I've updated the `if` clauses accordingly. Let me know whether this looks good 👍

byCedric: LGTM!

byCedric (approved):

PR #189, opened 2021-07-29
[website] Fix initial editor height

IjzerenHein:
# Why

Fixes the initial Monaco editor height. This problem started appearing in Chrome recently and has also been appearing in Firefox. Reported by @EvanBacon 

image

# How

- Use `react-resize-detector` to detect size changes

# Test Plan

- Verified on Chrome 92 locally (mac)
- Verified on Firefox 90 locally (mac)
- Verified on Safari 14.1.2 locally (mac)
- Verified on Edge 92 locally (mac)

### DIFF
--- a/website/src/client/components/shared/ResizeDetector.tsx
+++ b/website/src/client/components/shared/ResizeDetector.tsx
@@ -1,45 +1,19 @@
 import { StyleSheet, css } from 'aphrodite';
 import * as React from 'react';
+import { useResizeDetector } from 'react-resize-detector';
 
 type Props = {
-  onResize: () => void;
+  onResize: (width?: number | undefined, height?: number | undefined) => void;
   children: React.ReactNode;
 };
 
-export default class ResizeDetector extends React.Component<Props> {
-  componentDidMount() {
-    const horiz = this._horizontal.current;
-    const verti = this._vertical.current;
-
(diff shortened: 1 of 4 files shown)


website/src/client/components/shared/ResizeDetector.tsx
```diff
@@ -1,45 +1,19 @@
 import { StyleSheet, css } from 'aphrodite';
 import * as React from 'react';
+import { useResizeDetector } from 'react-resize-detector';
```
IjzerenHein, line 3: Depends on the ResizeObserver web API, which might not be supported on older browsers.
https://www.caniuse.com/resizeobserver

It's possible to use a polyfill as well.
https://www.npmjs.com/package/react-resize-detector

@byCedric What do you think?

byCedric: I think this is fine, +91% global support. The biggest non-supported browser would be IE, but I think that's fine since Microsoft dropped support for IE as well. If this becomes a problem, we can always add a "Hey, your browser is outdated" message.

Screenshot 2021-07-29 at 12 26 43

byCedric (approved):
Screenshot 2021-07-29 at 12 28 04

Nice!

PR #191, opened 2021-07-29
[website] make snack.expo.dev redirect permanent

FiberJW:
# Why

The main website and Snack have been migrated for a while now, we so should make this redirect permanent to signal which is the main website for SEO purposes.

# How

Changed response status to 308 and removed reversal code.


ide (commented):
Would it make sense to remove `isDevDomainEnabled()` too?

ide (approved):

PR #194, opened 2021-08-09
[ci] Add appetize workflow to check and deploy new simulator builds

byCedric:
# Why

We have almost all of the Snack development workflow in Github, except for Appetize. It's a manual step that requires some attention. This should fix that by adding a check & deploy workflow for Appetize.

# How

The workflow contains a couple of steps, with small differences for Android and iOS:

1. Resolve the current version on Appetize (main, embedded, or both queues)
2. Resolve the client version it _should have_ from SDK version
3. Download the client for the specified version
4. Prepare the client for Appetize (iOS only)
5. Upload the client to Appetize (main, embedded, or both queues)

With the inputs, we have a "double failsafe" to prevent undesired deployments. You both have to specify the exact SDK version to use, and you have to confirm with "deploy".

- When omitting both `deploy` and `sdkVersion`, only step 1 is executed.
- When omitting `deploy` but defining `sdkVersion`, steps 1 and 2 are executed.
- When defining both `deploy` and `sdkVersion`, all steps are executed.

# Required Environment variables

> ~~I haven't set these up yet!~~ Done, all set up! Thanks @tcdavis 

- `APPETIZE_MAIN_TOKEN` - account token for Appetize **main** queue
- `APPETIZE_MAIN_ANDROID` - public key for Appetize **main** Android app
- `APPETIZE_MAIN_IOS` - public key for Appetize **main** iOS app
- `APPETIZE_EMBED_TOKEN` - account token for Appetize **embed** queue
- `APPETIZE_EMBED_ANDROID` - public key for Appetize **embed** Android app
- `APPETIZE_EMBED_IOS` - public key for Appetize **embed** iOS app

# Test Plan

I created this workflow in my own repository, connected to my own Appetize instance. You [can see my test runs here](https://github.com/byCedric/snack-appetize-workflow). (_note, only `main` queue is set up_)

### DIFF
--- a/.github/workflows/appetize.yml
+++ b/.github/workflows/appetize.yml
@@ -1,0 +1,162 @@
+name: Appetize
+
+on:
+  workflow_dispatch:
+    inputs:
+      appetizeQueue:
+        description: Appetize queue to use (main/embed/all)
+        default: main
+        required: true
+      sdkPlatform:
+        description: Platform to use (android/ios/all)
+        default: all
+        required: true
+      sdkVersion:
+        description: SDK version to resolve Expo Go version (e.g. 42.0.0)
+        required: false
+      deploy:
+        description: type "deploy" to update Appetize (without this it only checks the version)
+        required: false
+
+jobs:
+  android:
+    if: contains('all android', github.event.inputs.sdkPlatform)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve Appetize (main) version
+        if: contains('all main', github.event.inputs.appetizeQueue)
+        # This step MUST only output `appVersionName`, the endpoint returns the public and private app keys
+        run: |
+          curl --silent -o appetize.json https://$APPETIZE_TOKEN@api.appetize.io/v1/apps/$APPETIZE_APP > /dev/null
+          echo "Appetize is running $(cat appetize.json | jq .appVersionName)"
+          rm -rf appetize.json
+        env:
+          APPETIZE_TOKEN: ${{ secrets.APPETIZE_MAIN_TOKEN }}
+          APPETIZE_APP: ${{ secrets.APPETIZE_MAIN_ANDROID }}
+      
+      - name: Resolve Appetize (embed) version
+        if: contains('all embed', github.event.inputs.appetizeQueue)
+        # This step MUST only output `appVersionName`, the endpoint returns the public and private app keys
+        run: |
+          curl --silent -o appetize.json https://$APPETIZE_TOKEN@api.appetize.io/v1/apps/$APPETIZE_APP > /dev/null
+          echo "Appetize is running $(cat appetize.json | jq .appVersionName)"
+          rm -rf appetize.json
+        env:
+          APPETIZE_TOKEN: ${{ secrets.APPETIZE_EMBED_TOKEN }}
+          APPETIZE_APP: ${{ secrets.APPETIZE_EMBED_ANDROID }}
+
+      - name: Resolve Android client
+        if: github.event.inputs.sdkVersion
+        id: android
+        run: |
+          curl --silent -o versions.json https://exp.host/--/api/v2/versions
+
+          VERSION=${{ github.event.inputs.sdkVersion }}
+          CLIENT_URL=$(cat versions.json | jq .sdkVersions[\"$VERSION\"].androidClientUrl)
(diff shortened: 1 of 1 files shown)


.github/workflows/appetize.yml
```diff
@@ -0,0 +1,162 @@
...
+      appetizeQueue:
+        description: Appetize queue to use (main/embed/all)
+        default: main
+        required: true
+      sdkPlatform:
+        description: Platform to use (android/ios/all)
+        default: all
+        required: true
+      sdkVersion:
+        description: SDK version to resolve Expo Go version (e.g. 42.0.0)
```
IjzerenHein, line 15: Would be great if we could somehow fill this in by default. Or when omitted that it retrieves the `sdkVersion` automatically, for instance from the `runtime` or `snack-sdk`

.github/workflows/appetize.yml
```diff
@@ -0,0 +1,162 @@
...
+          
+          if [ "${CLIENT_URL}" == "null" ]; then echo "No client url for SDK '$VERSION'"; exit 1; fi
+          if [ "${CLIENT_VERSION}" == "null" ]; then echo "No client version for SDK '$VERSION'"; exit 1; fi
+
+          echo "Resolved client $CLIENT_VERSION from SDK $VERSION"
+          echo "::set-output name=url::$CLIENT_URL"
+          echo "::set-output name=version::$CLIENT_VERSION"
+      
+      - name: Download Android client
+        if: github.event.inputs.deploy == 'deploy' && github.event.inputs.sdkVersion
```
IjzerenHein, line 66: Using `deploy` without filling in the `sdkVersion` will just skip this step. We should probably throw an error when using `deploy` while `sdkVersion` is not specified

.github/workflows/appetize.yml
```diff
@@ -0,0 +1,162 @@
...
+
+      - name: Upload to Appetize (embed)
+        if: contains('all embed', github.event.inputs.appetizeQueue) && github.event.inputs.deploy == 'deploy' && github.event.inputs.sdkVersion
+        # This step MUST NOT output anything, the endpoint returns the public and private app keys
+        run: |
+          curl --silent --http1.1 https://$APPETIZE_TOKEN@api.appetize.io/v1/apps/$APPETIZE_APP -F "file=@exponent-ios.zip" -F "platform=ios" > /dev/null
+          rm -rf exponent-ios.zip
+        env:
+          APPETIZE_TOKEN: ${{ secrets.APPETIZE_EMBED_TOKEN }}
+          APPETIZE_APP: ${{ secrets.APPETIZE_EMBED_IOS }}
```
IjzerenHein, line 162: Add a step to log to Slack, both when an error has occurred and when succesful

IjzerenHein (commented):
This is awesome 😍  😍  😍

.github/workflows/appetize.yml
byCedric: I think we can, but I'm not sure if that's a good idea. This workflow has the power of instantly breaking all appetizer emulators when using the wrong version. Forcing the initiator to think about the version it wants to use might be a good "sanity check" before deploying stuff. Not sure how you look at this though

.github/workflows/appetize.yml
byCedric: Yeah, maybe just erroring out might be better. I did this to avoid running this step because it actually did go off when SDK version is empty.

byCedric:
I'll try to refactor the android and ios client resolving, I don't like that I just copied and pasted it. Plus, it's a bit hard to find out what's going on.

.github/workflows/appetize.yml
byCedric: Done in the revised version #195

brentvatne (approved):

.github/workflows/appetize.yml
IjzerenHein: Okay fair enough. What will happen when you input an invalid version like "42.0.". Will this just fail with an understandable error?

IjzerenHein: Never mind, I can see from the code that it checks the version. LGTM

IjzerenHein (approved):

.github/workflows/appetize.yml
byCedric: Yes, it will error that the SDK 42.0. can't be found 😬

PR #195, opened 2021-08-11
[ci] Add appetize workflow to check and deploy new simulator builds

byCedric:
# Why

See #194, this is a full refactor of that version.

# How

See #194

# Test Plan

See #194

### DIFF
--- a/.github/workflows/appetize.yml
+++ b/.github/workflows/appetize.yml
@@ -19,144 +19,157 @@ on:
         required: false
 
 jobs:
+  prepare:
+    name: Preparing deploy
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.matrix.outputs.result }}
+    steps:
+      - name: 📋 Download SDK information
+        run: curl --silent -o versions.json https://exp.host/--/api/v2/versions
+      - name: 🔍 Resolve SDK versions
+        uses: actions/github-script@v4
+        if: github.event.inputs.sdkVersion
+        id: sdk
+        with:
+          script: |
+            const { sdkVersions } = require('./versions.json')
+            const version = '${{ github.event.inputs.sdkVersion }}'
+            const sdk = sdkVersions[version];
+
+            if (!sdk) throw new Error(`Could not find SDK "${version}"`)
+
+            core.info(`Resolved client versions for SDK "${version}"`)
+            core.info(`  - Android: ${sdk.androidClientVersion}`)
+            core.info(`  - iOS: ${sdk.iosClientVersion}`)
+
+            for (const key of ['androidClientVersion', 'androidClientUrl', 'iosClientUrl', 'iosClientVersion']) {
+              core.setOutput(key, sdk && sdk[key] ? sdk[key] : '')
+            }
+      - name: 👷‍♀️ Prepare tasks
+        uses: actions/github-script@v4
+        id: matrix
+        with:
+          script: |
+            const allQueues = ['main', 'embed']
+            const queue = '${{ github.event.inputs.appetizeQueue }}'
+
+            return {
+              include: allQueues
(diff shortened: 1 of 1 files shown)


.github/workflows/appetize.yml
```diff
@@ -19,144 +19,157 @@ on:
...
+            const version = '${{ github.event.inputs.sdkVersion }}'
+            const sdk = sdkVersions[version];
+
+            if (!sdk) throw new Error(`Could not find SDK "${version}"`)
+
+            core.info(`Resolved client versions for SDK "${version}"`)
+            core.info(`  - Android: ${sdk.androidClientVersion}`)
+            core.info(`  - iOS: ${sdk.iosClientVersion}`)
+
+            for (const key of ['androidClientVersion', 'androidClientUrl', 'iosClientUrl', 'iosClientVersion']) {
```
byCedric, line 46: Using `[...].forEach` kinda caused a race condition somehow. Not sure why, but you can [see it here](https://github.com/byCedric/snack-appetize-workflow/runs/3304590518?check_suite_focus=true#step:3:27).

.github/workflows/appetize.yml
```diff
@@ -19,144 +19,157 @@ on:
...
+      - name: 👷‍♀️ Prepare tasks
+        uses: actions/github-script@v4
+        id: matrix
+        with:
+          script: |
+            const allQueues = ['main', 'embed']
+            const queue = '${{ github.event.inputs.appetizeQueue }}'
+
+            return {
+              include: allQueues
```
byCedric, line 58: This is a nice trick to create a matrix set for both `android` and `ios`. This matrix is based on the different Appetize queues that we have. E.g. for `main` **OR** `embed`, only 1 item is created. For `all`, both `main` and `embed` are added. This results in a workflow like this:

Screenshot 2021-08-11 at 21 50 28

.github/workflows/appetize.yml
```diff
@@ -19,144 +19,157 @@ on:
...
         # This step MUST NOT output anything, the endpoint returns the public and private app keys
         run: |
-          curl --silent --http1.1 https://$APPETIZE_TOKEN@api.appetize.io/v1/apps/$APPETIZE_APP -F "file=@exponent-android.apk" -F "platform=android" > /dev/null
-          rm -rf exponent-android.apk
+          curl --silent --fail --http1.1 https://$TOKEN@api.appetize.io/v1/apps/$APP -F "file=@exponent-android.apk" -F "platform=android" > /dev/null
+          rm -rf exponent-android.apk appetize.json
         env:
-          APPETIZE_TOKEN: ${{ secrets.APPETIZE_MAIN_TOKEN }}
-          APPETIZE_APP: ${{ secrets.APPETIZE_MAIN_ANDROID }}
+          TOKEN: ${{ secrets[format('APPETIZE_{0}_TOKEN', matrix.appetize)] }}
```
byCedric, line 105: Nice handy trick to use "dynamic secrets through envvars".

byCedric:
Also, now when only using `deploy` (without sdkVersion), it just skips the version resolving. That makes the client download steps fail.

Screenshot 2021-08-11 at 21 52 52

.github/workflows/appetize.yml
brentvatne: this is a super weird one! it looks like the sdk object is somehow gc'd before we finish it.

.github/workflows/appetize.yml
brentvatne: clever

brentvatne (approved):

.github/workflows/appetize.yml
byCedric: Garbage collection seems to not work great in github scripts 😄

byCedric:
Merging this into the original PR.

PR #198, opened 2021-08-14
[website] Adds RudderStack to Snack

kgc00:
# Why

We are moving away from Segment in the near future and this PR introduces the alternative, RudderStack. For now, we will be keeping Segment in the codebase to ensure RudderStack is working as expected, then in a future PR we will remove Segment.

# How

- Adds the RudderStack client to `utils/Analytics.tsx`
- Adds the `RudderStackDocumentComponent`
- Adds env variables for RudderStack initialization to skaffold + dockerfile

One thing to note is that our current analytics for snack has prod and dev switched. For RudderStack, I kept sending prod events to the dev destinations but the RudderStack side properly labels the streams as dev/prod. Hopefully, at a future date, we can fix this issue.

# Test Plan

I loaded up a local version and fired some analytics events. Made sure that both Segment and RudderStack were firing and confirmed the events ended up in the proper destinations.


website/src/client/utils/Analytics.tsx
```diff
@@ -186,11 +194,13 @@ export default class Analytics {
       case 'identify':
         {
           const [traits, userId] = parameters;
-          if (typeof analytics !== 'undefined') {
+          if (typeof analytics !== 'undefined' && typeof rudderanalytics !== 'undefined') {
```
IjzerenHein, line 197: In case `rudderanalytics` could not be initialized, this would break recording to segment as well. Is this intentional?

IjzerenHein (changes requested):
Nice :))

The `RUDDERSTACK...` env vars also need to be added to `webpack.config.js` and the `deploy/staging` and `deploy/production` env files.

website/src/client/utils/Analytics.tsx
kgc00: > Nice :))
> 
> The `RUDDERSTACK...` env vars also need to be added to `webpack.config.js` and the `deploy/staging` and `deploy/production` env files.

Thanks! Yeah, I was missing the webpack config. re the env files - I do have the values updated per environment in the skaffold file, seems like this is also how the snack segment key sets its value. Happy to add in this change, but just curious, is putting it in the env files necessary or is it a style thing?

kgc00: > In case `rudderanalytics` could not be initialized, this would break recording to segment as well. Is this intentional?

Yeah, it was. There's this getter `isReady` and I was worried about it returning true, and losing some events, if rudder was initialized and segment wasn't or vice versa. But thanks for pointing this out, now that I think about it, it's more important to not break existing pipelines compared to the small number of events that we might lose from returning `isReady` before both analytics clients are initialized. Also, the scripts for rudder and segment both initialize an event queue before loading their respective libraries to store events until the instances are ready to handle them, so worrying about returning `isReady` too early is probably a non-issue. I'll be changing keep segment/rudder logic separate.

IjzerenHein: > Re the env files - I do have the values updated per environment in the skaffold file's docker build args. It seems like Segment's write key is set per environment in this way as well. I did add this in, but AFAIK the scaffold Dockerfile stuff is all that's necessary? just asking for my own edification 🙂

Very good question. Yeah the config is currently duplicated in both the `.env` files and the skaffold file. This still needs to be investigated on how to de-dupe these settings. The production/staging env files are currently accessed from `kustomization.yaml` that's why I've kept updating them. I'm not so experienced with skaffold and docker so I've been apprehensive in removing these files. Ideally we definitely do de-dupe these settings and probably mirror the setup from either `universe/www` or `universe/website`.

website/skaffold.yaml
```diff
@@ -20,6 +20,7 @@ build:
           DEPLOY_ENVIRONMENT: staging
           IMPORT_SERVER_URL: https://staging.snackager.expo.io
           LEGACY_SNACK_SERVER_URL: https://staging.snack.expo.io
+          RUDDERSTACK_WRITE_KEY: 1weTbQObpRMmIsWrdwEITRabLA2
```
IjzerenHein, line 23: Is this missing `RUDDERSTACK_DATA_PLANE_URL` ?

website/deploy/staging/snack.env
```diff
@@ -8,3 +8,5 @@ IMPORT_SERVER_URL=https://staging.snackager.expo.io
 DEPLOY_ENVIRONMENT=staging
 SNACK_WEBPLAYER_URL=https://snack-web-player-staging.s3.us-west-1.amazonaws.com
 SNACK_WEBPLAYER_CDN=https://d1qt8af2b3kxj0.cloudfront.net
+RUDDERSTACK_WRITE_KEY=1weTbQObpRMmIsWrdwEITRabLA2
+RUDDERSTACK_DATA_PLANE_URL=https://cdp.expo.dev
```
IjzerenHein, line 12: Missing newline

website/src/client/utils/Analytics.tsx
IjzerenHein: > Yeah, it was. There's this getter isReady, and I was worried about it returning true and losing some events when only one library was loaded... But thanks for pointing this out. Now that I think about it more thoroughly, it's more important to not break the existing pipeline compared to potentially losing some small number of events.
>
> Also, the scripts for rudder and segment both initialize an event queue to store events until the analytics instances are ready to handle them, so worrying about returning isReady too early is probably a non-issue. I'll be changing keep segment/rudder logic separate.

I do think that returning `isReady=true`  too eagerly is dangerous here. Imagine that `rudderanalytics` is initialized before segment `analytics` and then the `identify` queued event is processed. This would cause segment analytics to not receive the `identify` event and all subsequent track events would be identified incorrectly.

I think that in order to do this safely, we need multiple parallel pipelines. Each of them waiting for initialisation and they should never miss the 'identify` event. Probably refactor the code into a separate class (`AnalyticsProvider`?) which is initialized with a specific provider (segment,rudder,amplitude). And then have the top-level `Analytics` class, delegate the calls to all supported providers. Or something like that

website/deploy/production/snack.env
```diff
@@ -8,3 +8,5 @@ IMPORT_SERVER_URL=https://snackager.expo.io
 DEPLOY_ENVIRONMENT=production
 SNACK_WEBPLAYER_URL=https://snack-web-player.s3.us-west-1.amazonaws.com
 SNACK_WEBPLAYER_CDN=https://dwmszb351h4q.cloudfront.net
+RUDDERSTACK_WRITE_KEY=1weTlMb720s5AkBTYwgCGORCHHg
+RUDDERSTACK_DATA_PLANE_URL=https://cdp.expo.dev
```
IjzerenHein, line 12: Missing newline

website/src/client/utils/Analytics.tsx
kgc00: > I do think that returning isReady=true too eagerly is dangerous here. Imagine that rudderanalytics is initialized before segment analytics and then the identify queued event is processed. This would cause segment analytics to not receive the identify event and all subsequent track events would be identified incorrectly.

if you look at the scripts which load either RudderStack or Segment you'll see that both libraries instantiate an [event queue](https://github.com/expo/snack/pull/198/files#diff-616a0bc129c65a7b724f748b83dedd034af5e20afda88d0d620483f6cb7f2a25R27) immediately (similar to what you've done in the `Analytics` class). e.g. you'll see something like the following:
`var e=window.rudderanalytics=window.rudderanalytics||[];`. Then, they proceed to download their source code and instantiate an instance of their respective analytics classes. These libraries are loaded as a part of the top-level `Document` component, and therefore the event queues will be initialized before any analytics calls are made. For the `isReady` check to be true, either provider will need to finish downloading their source code and instantiate a class instance, which should take quite some time.

So, when an event like `analytics?.identify(userId, traits);` is fired, it adds that item to Segment's internal event queue to be processed when possible (remember, the queue exists even before `typeof analytics !== undefined` is true). Therefore, I don't think we really are in any danger here of calling analytics events too early and potentially losing events or not identifying users.

You can read more about how these libraries initialize themselves here: [RudderStack](https://github.com/expo/universe/blob/05d958a6766d28a38adf229331fb331a519c1290/server/website/common/analytics.tsx#L91) and [Segment](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/quickstart/#step-2-copy-the-segment-snippet). I personally think that the RudderStack docs do a better job of documenting what the code snippet is doing, so I recommend reading that one if you'd like to learn more about how these clients initialize. 

I think it's also worth noting that in the scripts to load Segment and RudderStack we include an identify call as soon as the library finished bootstrapping itself. IMO, this means that all users will be identified 100% of the time.

Happy to make the requested change but wanted to share some knowledge about these libraries because it seemed like we could get away with a simpler implementation.

website/skaffold.yaml
kgc00: yup, I added it in. In a recent update, I mistook `develop.env` for `base.env`

website/src/client/utils/Analytics.tsx
kgc00: So I switched back to requiring both analytics libraries to be initialized before `isReady` returns true. Sorry for the back and forth. After reading your comment I feel this is the best way forward.

Basically, we can be sure that both libraries will be loaded. They are embedded as `<script />` tags that run synchronously, and as soon as they execute they create event queues that store incoming events until the analytics clients finish initializing. You can read a description of how they work [here](https://docs.rudderstack.com/stream-sources/rudderstack-sdk-integration-guides/rudderstack-javascript-sdk#2-installing-the-rudderstack-javascript-sdk), but for our purposes, we can be sure that we will not lose events.

Also, we'll only be including Segment in the codebase for another week or two, so to rewrite the entire analytics class to support multiple providers doesn't seem to be a good option for me.

kgc00: > > Re the env files - I do have the values updated per environment in the skaffold file's docker build args. It seems like Segment's write key is set per environment in this way as well. I did add this in, but AFAIK the scaffold Dockerfile stuff is all that's necessary? just asking for my own edification 🙂
> 
> Very good question. Yeah the config is currently duplicated in both the `.env` files and the skaffold file. This still needs to be investigated on how to de-dupe these settings. The production/staging env files are currently accessed from `kustomization.yaml` that's why I've kept updating them. I'm not so experienced with skaffold and docker so I've been apprehensive in removing these files. Ideally we definitely do de-dupe these settings and probably mirror the setup from either `universe/www` or `universe/website`.

I see, thanks for the explanation! Yeah, when I implemented this for `universe/website` just using the docker args section of the skaffold was sufficient for deployments.

kgc00:
shouldn't this be `false`?
https://github.com/expo/snack/blob/4008b05ef142d4b3215f24b1b25890d4f790c217/website/src/client/utils/Analytics.tsx#L168

website/src/client/utils/Analytics.tsx
IjzerenHein: > Also, we'll only be including Segment in the codebase for another week or two, so to rewrite the entire analytics class to support multiple providers doesn't seem to be a good option for me.

Okay, yeah that is sound reasoning. I agree, there is no need to refactor into multiple pipelines if we are going to replace segment with rudderstack on short notice and this is just a temporary solution 👍 

> Basically, we can be sure that both libraries will be loaded. They are embedded as <script /> tags that run synchronously, and as soon as they execute they create event queues that store incoming events until the analytics clients finish initializing. You can read a description of how they work here (the Segment client works in the same fashion). It shouldn't really be necessary, but for our purposes, we can be sure that we will not lose events if we just wait for both. (I don't think there's ever a case where one client doesn't load at all).

Yeah I'm aware of this. The reason we added our own queuing was because the segment/amplitude initialisation/queuing wasn't stable. It sometimes ended up in an invalid state and we added our own queuing to guard against this, unfortunately..

IjzerenHein:
> shouldn't this be `false`?
> https://github.com/expo/snack/blob/4008b05ef142d4b3215f24b1b25890d4f790c217/website/src/client/utils/Analytics.tsx#L168

No I don't think so. The `isReady` check guards against an "incomplete" initialization of Amplitude. In development, there is neither segment, nor amplitude, and therefore `isReady=true` which causes the events to disappear in the void, instead of queueing up infinitely in the queue

IjzerenHein (approved):
LGTM

kgc00:
> > shouldn't this be `false`?
> > https://github.com/expo/snack/blob/4008b05ef142d4b3215f24b1b25890d4f790c217/website/src/client/utils/Analytics.tsx#L168
> 
> No I don't think so. The `isReady` check guards against an "incomplete" initialization of Amplitude. In development, there is neither segment, nor amplitude, and therefore `isReady=true` which causes the events to disappear in the void, instead of queueing up infinitely in the queue

gotcha. yeah, this function will always return true, then. might need a rewrite similar to:
but I won't worry about it here.

website/src/client/utils/Analytics.tsx
kgc00: gotcha! I did not think about Amplitude, and yeah this is all making more sense now 😅  - thanks for the explanation! 

I'm championing consolidating our analytics to RudderStack so we can have cross codebase consistency and a stable client that handles edge cases (like initialization and losing events) for us. Hopefully, in a future PR, we can remove Amplitude and simplify the whole setup 🙂 , but that's for another day.

IjzerenHein: Awesome that you are doing that, nice!!

When Amplitude is also replaced by RudderStack we should probably remove the whole queuing.

PR #200, opened 2021-08-23
update remaining relevant urls to expo.dev

FiberJW:
Forums, blog, emails, and the main website are now on expo.dev, so our code should reflect that.


IjzerenHein (approved):
Nice!

PR #204, opened 2021-08-27
[runtime] Add support for custom dev clients

esamelson:
# Why

ENG-1771

This PR adds support for custom dev clients to the snack runtime, via the properties added in https://github.com/expo/expo/pull/14195 (manifestURL constant export) and https://github.com/expo/expo/pull/14196 (disable auto-launch flag).

# How

- check for EXDevLauncher.manifestURL before falling back to Constants.getInitialURL if it doesn't exist.
- pass the new EXDevMenuDisableAutoLaunch intent/user defaults flag through appetize, in addition to the old one (EXKernelDisableNuxDefaults). This should be enough to prevent the dev menu from opening automatically when received by a dev client.

# Test Plan

I am not sure what the best way is to test this e2e -- I tried prebuilding the snack runtime but realized I need a server that will serve the correct format of snack manifest while also pointing to a local bundle. 🤔 

@byCedric do you think we can safely publish a new version of the snack runtime to staging in order to be table to test things e2e from there?


byCedric (approved):

website/src/client/utils/constructAppetizeURL.tsx
```diff
@@ -40,6 +40,7 @@ export default function constructAppetizeURL({
 
   const appetizeKey = constants.appetize.public_keys[previewQueue][platform];
   const appParams = {
+    EXDevMenuDisableAutoLaunch: true,
```
IjzerenHein, line 43: What does this do when using Expo Go on our default Appetize account? When is needed?

runtime/src/App.tsx
```diff
@@ -101,7 +102,7 @@ export default class App extends React.Component<object, State> {
     let initialURL: string | null = null;
     try {
       // Open from the initial URL if given
-      initialURL = await Linking.getInitialURL();
+      initialURL = EXDevLauncher.manifestURL ?? (await Linking.getInitialURL());
```
IjzerenHein, line 105: Didn't know you could combine async with non-async code like this. Nice :)

website/src/client/utils/constructAppetizeURL.tsx
byCedric: Correct me if I'm wrong @esamelson, but I don't think this does anything in Expo Go. It uses a different parameter to skip the welcome menu.

We need this when loading a devclient in Appetize, something we intend to do in the near future. This basically also skips the onboarding menu when opened for the first time.

PR #206, opened 2021-08-30
[website] Remove Segment analytics (was replaced by rudderstack)

kgc00:
# Why

Previously we introduced RudderStack as our Segment replacement. After comparing the event volumes between Segment and RudderStack we can see that RudderStack is functioning as expected and we can go ahead with removing Segment.

# How

I deleted all references to Segment from the codebase.

# Test Plan

I tested locally and confirmed events made it to the backend destinations.


IjzerenHein (approved):
Looks great!

PR #207, opened 2021-09-03
[snackager] Add motify org to forced reanimated plugin bundling

byCedric:
# Why

All of the `moti` / `@motify/` libraries require the reanimated plugin.

# How

- Added the org to the bundler
- Added cache busting to rebuild the libraries

# Test Plan

- See if staging works with [this snack](https://snack.expo.dev/@beatgig/courageous-peach)

### DIFF
--- a/snackager/src/cache-busting.ts
+++ b/snackager/src/cache-busting.ts
@@ -13,6 +13,8 @@ const cacheBusting: { version: number; packages: { [name: string]: number } } = 
   packages: {
     'react-native-reanimated': 1,
     moti: 1,
+    '@motify/interactions': 1,
+    '@motify/skeleton': 1,
     '@draftbit/ui': 1,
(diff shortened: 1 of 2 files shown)


snackager/src/cache-busting.ts
```diff
@@ -13,6 +13,8 @@ const cacheBusting: { version: number; packages: { [name: string]: number } } =
   packages: {
     'react-native-reanimated': 1,
     moti: 1,
+    '@motify/interactions': 1,
+    '@motify/skeleton': 1,
```
IjzerenHein, line 17: You can also use wildcards to invalidate all `@motify` packages, like this:

```
'@motify/.*': 1,
```

IjzerenHein: @byCedric

Issue #209, opened 2021-09-06
[website] ResizeObserver loop limit exceeded


IjzerenHein:
### Summary

![image](https://user-images.githubusercontent.com/6184593/132207579-2030a2c6-8f1f-42ec-a8f3-b7a3d2ef02bc.png)

app.bundle.js:109 [Util] handleError::  ResizeObserver loop limit exceeded
13:01:23.435 app.bundle.js:109 [Util] handleError::  ResizeObserver loop limit exceeded
13:01:26.985 app.bundle.js:109 [Util] handleError::  ResizeObserver loop limit exceeded

Background info here: https://github.com/maslianok/react-resize-detector/issues/45

### What platform(s) does this occur on?

Android, iOS, Web

### SDK Version

42

### Reproducible demo or steps to reproduce from a blank project

To reproduce:

The error does not show up in development by default, but is detected by segment/rudderstack.
To visualizer the error:

- Add `window?.addEventListener('error', function (e) { console.error(e); });` somewhere to the `website` code
- `yarn start` (in root)
- Resize the browser

Happened on this Snack, but could have been any I think:
https://snack.expo.dev/@hein_expo/shared-element-2?verbose=true

Preview pane open: MyDevice
Browser: Chrome (mac)

IjzerenHein:
Fixed by f26d6befbcc1f22579c054c7f8302e4f6aa01e41

PR #210, opened 2021-09-09
[website] Disable Sentry error reporting to avoid error spam

byCedric:
# Why

The website had an incident where it spammed Sentry quite a lot. Let's disable Sentry on the website until we can find a way to separate website issues from user issues.

# How

Disabled the install from code, and disabled the Sentry envvar.

# Test Plan

- Check is staging works


IjzerenHein (approved):
Some work has already been done in these PRs #201 and #202.

Many of the errors remaining and related to the Monaco editor, which requires an upgrade and more investigation. The upgrade is blocked though due to various dependencies and because we still use webpack@4.
And there is also this issue  #209

Not sure how to deal with all the `Operation timed out` issues those. Should these service errors not be logged?

byCedric:
> Not sure how to deal with all the `Operation timed out` issues those. Should these service errors not be logged?

I think not-reporting them would be good for now. This is not necessarily a bug or issue in the code, it's related to performance. That might be better for analytics instead of Sentry?

PR #216, opened 2021-09-17
[website] do not dismiss modal on content overselect

Simek:
# Why

Currently, when user overselect the input or other content inside Modal, it automatically closes, which could be a bit annoying.

This is because the click started inside the modal ends outside of it, and click ends on the Modal overlay.

Kapture 2021-09-17 at 15 16 20

# How

To prevent this behaviour I have swapped the `onClick` listener to `onMouseDown` for the modal overlay component. 

# Test Plan

The change has been tested running the Snack website locally.

# Preview

Kapture 2021-09-17 at 15 09 51

### DIFF
--- a/website/src/client/components/shared/Modal.tsx
+++ b/website/src/client/components/shared/Modal.tsx
@@ -75,7 +75,7 @@ export default class Modal extends React.PureComponent<Props, State> {
           !this.state.initial && styles.initial,
           this.props.visible ? styles.visible : styles.hidden
         )}
-        onClick={this._handleDismiss}>
+        onMouseDown={this._handleDismiss}>
         <div ref={this._content} className={css(styles.content)}>
           {this.state.rendered ? this.props.children : null}
         </div>


IjzerenHein (commented):
Oh I know what you mean, it's so annoying that it sometimes closes the modal because of that. Thanks for looking into this!

Have you also verified it works on mobile (eg tablet)?

Simek:
> Have you also verified it works on mobile (eg tablet)?

Initially I have tested this in Chrome via device emulation (with touch events enabled), a while ago I have also verified this using Simulator. In both cases I have used iPad Pro. 🙂

IjzerenHein:
> > Have you also verified it works on mobile (eg tablet)?
> 
> Initially I have tested this in Chrome via device emulation (with touch events enabled), a while ago I have also verified this using Simulator. In both cases I have used iPad Pro and everything seems to be fine. 🙂

Okay awesome, thanks for verifying this @Simek !

IjzerenHein (approved):

PR #220, opened 2021-09-28
[website] Properly initialize Amplitude to include referrals

byCedric:
# Why

Right now, it doesn't capture all referrals.

# How

See https://developers.amplitude.com/docs/javascript#track-referrers

# Test Plan

Deploy and see if it improves.


IjzerenHein:
Nice!

PR #225, opened 2021-10-08
[runtime] Add expo random as dependency

byCedric:
# Why

This was causing some issues when creating a dev-client build of the runtime. We are using it in https://github.com/expo/snack/blob/main/runtime/src/NativeModules/getDeviceIdAsync.tsx#L2

# How

- Run `expo install expo-random` in runtime

# Test Plan

- See if CI passes


byCedric:
Will recreate to avoid the merge conflict (and use the proper version)

PR #240, opened 2021-12-09
[website] store editor theme and mode in local storage

Simek:
# Why

Currently, often first thing done when I open a new Snack is to change the theme to dark mode. 😅 

Since the editor preferences are non-crucial or sensitive data in any way we can use the browser local storage to store them to improve the users UX.

# How

This PR adds store and load code for the editor theme and mode.

I have placed this on top of all overrides for it has the less priority and will be overwritten by, for example, theme provided in query string.

Also to improve the readability a bit I have extracted the editor mode possible values to the separate type.

# Test Plan

The changes have been tested by running the website code on localhost.

# Preview

> No preferences in LS => Set/toggle preferences => Reload website

Kapture 2021-12-09 at 16 24 09

### DIFF
--- a/website/src/client/components/Preferences/PreferencesProvider.tsx
+++ b/website/src/client/components/Preferences/PreferencesProvider.tsx
@@ -9,12 +9,13 @@ import type { PreferencesContextType } from './withPreferences';
 import type { ThemeName } from './withThemeName';
 
 export type PanelType = 'errors' | 'logs';
+export type EditorModeType = 'normal' | 'vim';
 
 export type PreferencesType = {
   deviceConnectionMethod: ConnectionMethod;
   devicePreviewPlatform: Platform;
   devicePreviewShown: boolean;
-  editorMode: 'normal' | 'vim';
+  editorMode: EditorModeType;
   fileTreeShown: boolean;
   panelsShown: boolean;
   panelType: PanelType;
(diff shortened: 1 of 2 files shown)


byCedric (approved):
Thanks 😄

website/src/client/components/Preferences/PreferencesProvider.tsx
```diff
@@ -61,6 +62,16 @@ class PreferencesProvider extends React.Component<Props, State> {
 
     let overrides: Partial<PreferencesType> = {};
 
+    try {
+      // Restore editor preferences from local storage
+      if (localStorage) {
+        overrides.editorMode = localStorage.getItem('editorMode') as EditorModeType;
+        overrides.theme = localStorage.getItem('theme') as ThemeName;
```
byCedric, line 69: @Simek FYI, I'll add some defaults to this check. Right now it uses the wrong theme when users did not store any data yet. (wrong syntax highlighting)

Simek: Oh, that's a bit surprising to me, but thanks for catching this. 👍 

I have tested similar scenario when I was checking that query params overwrites the local storage values IIRC, but probably I have missed this issue somehow. 😅

PR #250, opened 2022-01-20
[website] Change settings link text to match main website

wschurman:
# Why

The main expo.dev website now calls these settings "User Settings" and refers to them as such in the similar dropdown on the website:

Screen Shot 2022-01-20 at 12 17 00 PM

# How

Grep, change text.

# Test Plan

Inspect.


wschurman:
I'm also open to just removing the link to settings here. It doesn't actually make that much sense to me to have it here.

jonsamp (approved):
I like this change for consistency

PR #251, opened 2022-02-12
[snack-content] Split out code related to Snack project definitions into `snack-content`

ide:
Why
---
The Expo Snack server uses `snack-sdk` for some functions and type definitions, but it doesn't need client-side code like transports. I chose the name `snack-content` over `snack-projects` to avoid autocomplete conflicts with `snack-proxies`. Closes ENG-3910

Commit 1: Split out code into `snack-content` 
Commit 2: Make `snack-sdk` use `snack-content`
Commit 3: GHA

How
---
The functions in sdk.ts and about half of the type definitions were moved from `snack-sdk` to a new package called `snack-content`, which is also a workspace package. I also changed the tsconfig target to ES2018 which assumes support for Object.assign and object rest spread syntax. GitHub is requiring ES2020, so ES2018 is relatively conservative for a developer-heavy userbase.

The public API functions copied over are: isModulePreloaded, getPreloadedModules, validateSDKVersion, getDeprecatedModule, getSupportedSDKVersions, isValidSemver, isFeatureSupported, standardizeDependencies. The public API types copied over are: SnackFile, SnackCodeFile, SnackAssetFile, SnackFiles, SnackDependencyVersions, SnackDependency, SnackDependencies, SnackMissingDependency, SnackMissingDependencies, SDKVersion, and SDKFeature.

The functions and types that moved to `snack-content` have been removed from `snack-sdk`, which now imports `snack-content`. There should be no API changes for `snack-sdk` except that it requires ES2018 (about four years old at this point).

Test Plan
--------
Moved over the relevant unit tests from `snack-sdk` into `snack-content`. Added more unit tests for nearly full coverage. Unit tests for `snack-content` pass.


packages/snack-content/src/sdks/types.ts
```diff
@@ -40,10 +40,4 @@ export type SDKSpec = {
...
  * Feature that is supported by the SDK (e.g. "TYPESCRIPT").
  */
-export type SDKFeature =
-  | 'MULTIPLE_FILES'
-  | 'PROJECT_DEPENDENCIES'
-  | 'TYPESCRIPT'
-  | 'UNIMODULE_IMPORTS'
-  | 'POSTMESSAGE_TRANSPORT'
-  | 'VERSIONED_SNACKAGER';
+export type SDKFeature = 'TYPESCRIPT';
```
ide, line 43: I kept this since it's used in tests

packages/snack-content/src/sdk.ts
```diff
@@ -133,3 +108,28 @@ export function standardizeDependencies(dependencies: any): SnackDependencies {
...
+      result[name] = peerDep.version;
+    } else {
+      // Invalid peer-dependency
+      result = result === peerDependencies ? { ...peerDependencies } : result;
+      delete result[name];
+    }
+  }
+
+  return result;
+}
```
ide, line 135: No code changes here -- moved it below standardizeDependencies so that the more important and public functions come first, and implementation details come after.

byCedric: It's also not the easiest function to understand, thanks for leaving it intact 😄

packages/snack-content/src/sdks/types.ts
byCedric:
```suggestion
// TODO(cedric): refactor tests to use types from `snack-content`
export type SDKFeature = 'TYPESCRIPT';
```

byCedric (approved):
Looks great! 

I think one last thing we need to do before being able to deploy this is updating the Dockerfiles for [website](https://github.com/expo/snack/blob/main/website/Dockerfile) and [snackager](https://github.com/expo/snack/blob/main/snackager/Dockerfile). It manually copies over a couple of folders to build from. 

Might just copy the full repo in the "dev" intermediate container to not have to think about this in the future.

ide:
I'm working on updating the Dockerfiles. Using .dockerignore we can make it easier to copy workspace packages into the Docker image but we will still need to add a build step for each package.

PR #259, opened 2022-04-05
[snackager] Upgrade Node from 12 to 16

danstepanov:
https://linear.app/expo/issue/ENG-4551/update-snackager-to-node-16-lts-v16142

[snackager] Upgrade Node from 12 to 16

Closes #ENG-4551

# Why
There is no reason for snackager to be running such an old version of node.

# How
- Update snackager to run node 16.14.2, via Volta. 
- Beyond the explicit node version bump, skaffold uses an updated method of calling alpine. 
- Github Actions reference node 16 for proper cloud deployment. With the bump to node 16.14.2, typescript is updated. 
- There is no way to guarantee that an object thrown in a catch clause is a Error subclass. Subsequently,
allow changing the type of the variable from `any` to `unknown`.
- Integration tests reflect changes made to fs functions since node 12.18.4.
- Modularize building snackager such that snackager and its dependencies can all have varying node and typescript versions. Update to snackager Github Action to manually build dependencies. These changes will be applied to the rest of the repo over time.
- Downgrade npm to stable 6.14.16 to allow e2e tests to pass.
- This downgrade needs to be
run outside of the repo's folder, rearrange CI workflows, accordingly.

# Test Plan
- Tested with `skaffold dev -f snackager/skaffold.yaml`, run from the root of the repository. Please note that if you are using Docker Desktop, you will need to change line 44 of `snackager/skaffold.yaml` to `docker-desktop` rather than `minikube`. Docker Desktop should generally be used but this will be reflected in a future commit.


byCedric:
I think you just found a small issue caused by upgrading to Node 16 😄  It seems that we need to update the logic for this failing test case. [You can see this should have passed on the current `main` here](https://github.com/expo/snack/runs/5772516793?check_suite_focus=true#step:15:25). Although this would be a slightly late commit, let's try and resolve that before merging this!

snackager/package.json
```diff
@@ -82,7 +82,7 @@
     "lint": "eslint ./src"
   },
   "volta": {
-    "node": "12.18.4",
+    "node": "16",
```
byCedric, line 85: I would put a slightly more specific version here for volta, [like the current latest 16 version](https://nodejs.org/download/release/latest-v16.x/): `16.14.2`

snackager/skaffold.yaml
```diff
@@ -12,7 +12,7 @@ build:
     docker:
       dockerfile: snackager/Dockerfile
       buildArgs:
-        node_version: 12.18.4
+        node_version: 16
```
byCedric, line 15: Same here. I would put an exact version here instead.

packages/snack-sdk/src/DependencyResolver.ts
```diff
@@ -112,7 +112,7 @@ export default class DependencyResolver {
         fileName: key,
       });
       this.logger?.error(error);
-      if (this.status[key]) {
+      if (key) {
```
byCedric, line 115: This seems like a behavioral change, can you give some context on why this was necessary?

snackager/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN yarn build
 
 # Minimize image
 RUN yarn install --frozen-lockfile --offline --production
-FROM mhart/alpine-node:${node_version}
+FROM node:${node_version}-alpine
```
byCedric, line 52: Nick would be _very_ happy 😄

byCedric (approved):
Small nitpicks related to whitespaces, only 1 bigger thing: https://github.com/expo/snack/pull/259#discussion_r845474838

After that I think it's good to go!

packages/snack-sdk/src/DependencyResolver.ts
danstepanov: `this.status[key]` always returns a Promise, which causes a type error. As all that is being done in that if statement is deleting the relevant item for the key and, since the key is of type `string`, you can simply check for the key instead of checking for the item for said key.

danstepanov: if left as a Promise, there will both be a type error and it will always return `true`

byCedric: I'm not sure if I follow, the `this.status` dictionary should contain the (latest?) promises related to dependency actions. So I guess it's an important part of checking if something is going on right?

We have this check in multiple places, e.g. https://github.com/expo/snack/blob/main/packages/snack-sdk/src/DependencyResolver.ts#L76-L86. And if there is no pending action going on, then it shouldn't try to delete it right?

I agree on the "all that is being done is deleting the relevant status", but I'm not sure if that always will be like that. Plus it does log something, which means it could trigger a log while that log it isn't relevant, right? That log might confuse users maybe?

PR #270, opened 2022-04-13
[babel-standalone] add babel standalone to snack monorepo

byCedric:
# Why

This should add some visibility on the exact presets and babel versions used in the snack runtime.

> Note, this package has already been published in its current state

# How

- Copied content from https://github.com/expo/babel-standalone
- Renamed package to `snack-babel-standalone`
- Decoupled the package from the monorepo in order to avoid multiple babel versions being compiled (the side goes from ±25mb to ±75mb if we dont do this).

# Todo (after merge in follow-up PR)

- Update the references from `@expo/babel-standalone` to `snack-babel-standalone`
- Update the babel instances in `snack-babel-standalone` to latest versions
- Add additional plugins and presets like `@babel/plugin-transform-async-to-generator` (for using JSI host objects in async functions).

# Test Plan

TBD


packages/snack-babel-standalone/.editorconfig
```diff
@@ -0,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
danstepanov, line 1: is this different than eslint?

danstepanov (approved):

packages/snack-babel-standalone/.editorconfig
byCedric: editorconfig is basically a settings definition for your IDE. And it's supported by many, like vscode, jetbrains, vim, etc (usually through an additional plugin). So eslint lints code, editorconfig sets your IDE.

PR #274, opened 2022-04-14
Update website to Node 16

danstepanov:
# Why

We want all of snack to be using Node 16 over 12

# How

* [website] Update CI to use Node 16

* [website] Bump to Node 16.14.2

* [website] Remove TS reference to snack-sdk and snack-content

* [website] Update skaffold to reflect node 16.14.2

# Test Plan

TBD

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -16,14 +16,15 @@
     "clean": "yarn clean:server && yarn clean:client",
     "clean:server": "del build/",
     "clean:client": "del dist/",
-    "lint": "eslint ./src ./typings",
+    "lint": "eslint ./src",
     "test": "env-cmd -e test jest --config jest/unit.config.js",
     "prebuild": "yarn clean"
   },
   "author": "Expo",
   "license": "MIT",
   "volta": {
-    "node": "12.18.4"
+    "node": "16.14.2",
+    "npm": "6.14.16"
   },
   "dependencies": {
     "@babel/parser": "^7.12.16",
@@ -89,7 +90,7 @@
     "@babel/preset-env": "^7.12.16",
     "@babel/preset-react": "^7.12.13",
     "@babel/preset-typescript": "^7.12.16",
-    "@tsconfig/node12": "^1.0.7",
+    "@tsconfig/node16": "^1.0.2",
     "@types/amplitude-js": "^6.0.0",
     "@types/aphrodite": "^0.5.13",
     "@types/bunyan": "^1.8.6",
@@ -157,7 +158,7 @@
     "ts-jest": "^26.5.0",
     "ts-node-dev": "^1.1.1",
     "tsconfig-paths": "^3.9.0",
-    "typescript": "^4.1.3",
+    "typescript": "^4.6.3",
     "webpack": "^4.29.0",
     "webpack-cli": "^3.2.1",
     "webpack-dev-middleware": "^3.5.1",
--- a/website/src/client/components/Import/ImportFilesManager.tsx
+++ b/website/src/client/components/Import/ImportFilesManager.tsx
@@ -21,7 +21,7 @@ type Props = {
   className?: string;
 };
 
-type FileItem = File | WebkitFileEntry | WebkitDirectoryEntry;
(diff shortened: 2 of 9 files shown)


website/package.json
```diff
@@ -76,6 +77,7 @@
     "recast": "^0.20.3",
     "redux": "^4.0.1",
     "sanitize-html": "^1.20.0",
+    "snack-content": "*",
```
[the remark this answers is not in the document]
danstepanov: I'll also remove this from CI then

byCedric:
Looking good. Almost ready to merge! Just the TS issues reported in CI 😄

website/src/client/components/Import/ImportFilesManager.tsx
```diff
@@ -136,7 +138,7 @@ export default class ImportFilesManager extends React.PureComponent<Props, State
         folderMappings[item.name] = getUniquePath(allPaths, item.name);
       }
     });
-
+    // @ts-ignore
```
byCedric, line 141: What's the error you get for this one?

website/src/client/components/Import/ImportFilesManager.tsx
```diff
@@ -87,11 +87,13 @@ export default class ImportFilesManager extends React.PureComponent<Props, State
...
-          const file = Array.from(dataTransfer.files).find((file) => file.name === entry.name);
-          return entry.isFile && file ? file : entry;
-        });
+        entries = Array.from(dataTransfer.items)
+          .map((it) => it.webkitGetAsEntry())
+          .filter((entry): entry is globalThis.FileSystemEntry => entry != null)
+          .map((entry) => {
+            const file = Array.from(dataTransfer.files).find((file) => file.name === entry.name);
+            return entry.isFile && file ? file : entry;
+          });
```
byCedric, line 96: This makes sense, not sure in what cases `it.webkitGetAsEntry` would return null though. The only thing I would change is this:

```suggestion
        entries = Array.from(dataTransfer.items)
          .map((it) => it.webkitGetAsEntry())
          .filter((entry): entry is FileSystemEntry => entry != null)
          .map((entry) => {
            const file = Array.from(dataTransfer.files).find((file) => file.name === entry.name);
            return entry.isFile && file ? file : entry;
          });
```

website/src/client/components/Import/ImportFilesManager.tsx
```diff
@@ -21,7 +21,7 @@ type Props = {
   className?: string;
 };
 
-type FileItem = File | WebkitFileEntry | WebkitDirectoryEntry;
+type FileItem = File | WebkitFileEntry | WebkitDirectoryEntry | globalThis.FileSystemEntry;
```
byCedric, line 24: I don't think this one is required, right? The entries aren't `FileItem`s or `FileSystemEntry`'s, it's something that's being used after everything is processed.

danstepanov: without this added type, there is a typescipt error [here](https://github.com/expo/snack/blob/ad8634a384f5f586af2ddf89effa19483dbc1cb3/website/src/client/components/Import/ImportFilesManager.tsx#L90). The array has both `File` and `FileSystemEntry` items.
Screen Shot 2022-04-15 at 3 56 33 PM

website/src/client/components/Import/ImportFilesManager.tsx
danstepanov: it seems like `globalThis.FileSystemEntry` and `FileSystemEntry` may be different types. when making your requested change, typescript complains about type `FileSystemEntry` not being assignable to `FileSystemEntry | null`. I'm also not sure why this value might be null.

website/src/client/components/Import/ImportFilesManager.tsx
danstepanov: Screen Shot 2022-04-15 at 4 03 02 PM

byCedric (approved):
Alright, changes make sense 😄 thanks for the clarifications

PR #282, opened 2022-05-04
[snackager] Add missing text input external for Stripe

byCedric:
# Why

Stripe uses this one, so we need to mark it as external.

# How

Would be better if we can use the package-spec instead. e.g.

- `react-native/*` → external because of `react-native`

# Test Plan

- Check if `@stripe/stripe-react-native` builds with Snackager
> [There is an ongoing issue with the files pointing to a non-existing `package.json`. Once this is resolved, we can test this](https://github.com/stripe/stripe-react-native/pull/924).


byCedric:
`@stripe/stripe-react-native@0.9.0` [built successfully](https://github.com/expo/snack/runs/6379219361?check_suite_focus=true#step:6:240), rolling out this fix.

PR #289, opened 2022-05-05
[CI] End rollouts to old cluster

nicknovitski:
# Why

All traffic has been switched to the services running in the general-central cluster.


byCedric:
I can't comment on non-edited code, but there are references to `🚀 Deploy x in old cluster` ([this for example](https://github.com/expo/snack/blob/b5bba2babccca19e170c9be35155e204c8742602/.github/workflows/snackager.yml#L130-L132))

Might be good to rename that to just `🚀 Deploy x` 😄  Other than that, LGTM!

byCedric (approved):
see comment!

PR #292, opened 2022-05-18
test commit

danstepanov:
# Why

# How

# Test Plan


byCedric:
Before we can merge this, I'm checking with the SDK team why `react-native-gesture-handler` was not upgraded during the SDK 45 cycle. We still use `~2.2.1` in Expo Go, adding this will likely cause the native layer to be out of sync.

himrocks33:
This does fix the issue on web but it introduces a new issue on android: 
`["RNGestureHandlerRootView" was not found in the UIManager]`
I attempted suggested fixes here:
https://stackoverflow.com/questions/71396210/rngesturehandlerrootview-was-not-found-in-the-uimanager
None of which seem to make a difference. Only way to get android to work is to downgrade RN Gesture Handler.

PR #297, opened 2022-05-20
[website] Set status code for not found responses to avoid CDN caching

byCedric:
# Why

Fixes [ENG-5064](https://linear.app/expo/issue/ENG-5064/fix-caching-headers-for-snacks-static-assets-and-404s)

Our CDN cached a 404 response for some chunks of assets, causing Snack to break until the CDN was flushed. This should help tell the CDN to not-cache 404 responses.

# How

Snack's website is a kind of custom-built, a legacy from how it was created years ago. So the implementation is not ideal, but here is how it works:
- Koa tries to serve static files first. If that doesn't exist, it continues through the other middlewares.
- When no static files are found, it server-side renders the **src/client/components/Router.tsx** 
- If that client router is being rendered server-side, pass the server context.
- Whenever the client router needs to respond with a not-found page, set the right status and header (if available, so if server-side rendered).

The unfortunate part here is that we need to "leak" the server context to the client-side router. We have no other way to determine if a 404 hit was rendered since we only get the HTML output from that.

# Test Plan

Run it locally, or deploy to staging and try the following URLs:
- https://staging.snack.expo.dev/this-path-doesnt-exist
- https://staging.snack.expo.dev/@bycedricstaging/this-snack-doesnt-exist
- https://staging.snack.expo.dev/dist/0.3babf018b826ac329594.chunk.FAKE.js

Either one of these URLs should return with a 404 status code, and the right cache headers.

Optionally, [test the URLs from the client router](https://github.com/expo/snack/blob/main/website/src/client/components/Router.tsx#L37-L42), these should respond with 200 status codes.

```jsx
<Route exact path="/embedded/@:username/:projectName+" render={renderRoute} />
<Route exact path="/embedded/:id" render={renderRoute} />
<Route exact path="/embedded" render={renderRoute} />
<Route exact path="/@:username/:projectName+" render={renderRoute} />
<Route exact path="/:id" render={renderRoute} />
<Route exact path="/" render={renderRoute} />
```

### DIFF
--- a/website/src/client/components/Router.tsx
+++ b/website/src/client/components/Router.tsx
@@ -28,7 +28,7 @@ export default function Router(props: RouterProps) {
     }
 
     return <NonExistent />;
-  }, []);
+  }, [props.ctx]);
 
   const renderRoute = React.useCallback(
     (routeProps: any) => {
--- a/website/src/client/components/Router.tsx
+++ b/website/src/client/components/Router.tsx
@@ -18,8 +18,8 @@ type RouterProps = {
 export default function Router(props: RouterProps) {
   const render404 = React.useCallback(() => {
     if (props.ctx) {
-      // If te router is server-side rendered, we need to set the right response code and headers.
-      // Without this, possible CDNs might cache the invalid path.
+      // If the router is server-side rendered, we need to set the right response code and headers.
+      // Without this, it's possible CDNs might cache the invalid path.
       props.ctx.response.status = 404;
       props.ctx.response.set(
         'Cache-Control',
--- a/website/src/client/components/Router.tsx
+++ b/website/src/client/components/Router.tsx
@@ -30,27 +30,24 @@ export default function Router(props: RouterProps) {
     return <NonExistent />;
   }, [props.ctx]);
 
-  const renderRoute = React.useCallback(
-    (routeProps: any) => {
-      const { data, ...rest } = props;
-      const isEmbedded = routeProps.location.pathname.split('/')[1] === 'embedded';
+  function renderRoute(routeProps: any) {
+    const { data } = props;
+    const isEmbedded = routeProps.location.pathname.split('/')[1] === 'embedded';
 
-      if (!data || data.type !== 'success') {
-        return render404();
-      }
+    if (!data || data.type !== 'success') {
+      return render404();
+    }
 
-      const appProps = {
-        ...routeProps,
-        ...rest,
-        query: props.queryParams,
-        snack: data.snack,
-        defaults: data.defaults,
-      };
+    const appProps: React.ComponentProps<typeof App | typeof EmbeddedApp> = {
+      ...routeProps,
+      userAgent: props.userAgent,
+      query: props.queryParams,
+      snack: data.snack,
+      defaults: data.defaults,
+    };
 
-      return isEmbedded ? <EmbeddedApp {...appProps} /> : <App {...appProps} />;
-    },
-    [render404]
-  );
+    return isEmbedded ? <EmbeddedApp {...appProps} /> : <App  {...appProps} />;
+  }
 
   return (
     <Switch>
--- a/website/src/client/components/Router.tsx
+++ b/website/src/client/components/Router.tsx
@@ -46,7 +46,7 @@ export default function Router(props: RouterProps) {
       defaults: data.defaults,
     };
 
-    return isEmbedded ? <EmbeddedApp {...appProps} /> : <App  {...appProps} />;
+    return isEmbedded ? <EmbeddedApp {...appProps} /> : <App {...appProps} />;
   }
 
   return (
--- a/website/src/client/components/Router.tsx
+++ b/website/src/client/components/Router.tsx
@@ -21,31 +21,36 @@ export default function Router(props: RouterProps) {
       // If te router is server-side rendered, we need to set the right response code and headers.
       // Without this, possible CDNs might cache the invalid path.
       props.ctx.response.status = 404;
-      props.ctx.response.set('Cache-Control', 'private, no-cache, no-store, max-age=0, must-revalidate');
+      props.ctx.response.set(
+        'Cache-Control',
+        'private, no-cache, no-store, max-age=0, must-revalidate'
+      );
     }
 
     return <NonExistent />;
   }, []);
 
-  const renderRoute = React.useCallback((routeProps: any) => {
-    const { data, ...rest } = props;
-    const isEmbedded = routeProps.location.pathname.split('/')[1] === 'embedded';
+  const renderRoute = React.useCallback(
+    (routeProps: any) => {
+      const { data, ...rest } = props;
+      const isEmbedded = routeProps.location.pathname.split('/')[1] === 'embedded';
 
-    if (!data || data.type !== 'success') {
-      return render404();
-    }
+      if (!data || data.type !== 'success') {
+        return render404();
+      }
 
-    const appProps = {
-      ...routeProps,
-      ...rest,
-      query: props.queryParams,
-      snack: data.snack,
-      defaults: data.defaults,
-    };
-      
-    return isEmbedded ? <EmbeddedApp {...appProps} /> : <App {...appProps} />;
-  }, [render404]);
-  
+      const appProps = {
+        ...routeProps,
+        ...rest,
+        query: props.queryParams,
+        snack: data.snack,
+        defaults: data.defaults,
+      };
+
+      return isEmbedded ? <EmbeddedApp {...appProps} /> : <App {...appProps} />;
+    },
+    [render404]
+  );
 
   return (
     <Switch>
--- a/website/src/client/components/Router.tsx
+++ b/website/src/client/components/Router.tsx
@@ -1,3 +1,4 @@
+import { Context as ServerContext } from 'koa';
 import * as React from 'react';
 import { Switch, Route } from 'react-router-dom';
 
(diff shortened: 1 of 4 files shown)


byCedric:
I did not add any cache headers to the valid assets being served by `koa-static`. Currently, I'm not certain if adding that would not conflict with the service worker's cache. We've had a lot of issues in the past with that, so I prefer not to touch that one if that's ok. @ide

ide:
Thanks for working on this so quickly!

> We've had a lot of issues in the past with that, so I prefer not to touch that one if that's ok.

Yes, I think that's OK for now and should be a separate PR if we decide to investigate it. We would need to make sure to apply the caching headers only to URLs that change whenever the content changes (URLs that contain a checksum of the content, or URLs that change on every deployment). We don't want to cache URLs that can change between deployments.

website/src/client/components/Router.tsx
```diff
@@ -6,42 +7,60 @@ import App from './App';
...
 
-export default class Router extends React.Component<Props> {
-  private _renderRoute = (props: any) => {
-    const { data, ...rest } = this.props;
-    const isEmbedded = props.location.pathname.split('/')[1] === 'embedded';
+export default function Router(props: RouterProps) {
+  const render404 = React.useCallback(() => {
+    if (props.ctx) {
+      // If te router is server-side rendered, we need to set the right response code and headers.
+      // Without this, possible CDNs might cache the invalid path.
```
ide, line 22:
```suggestion
      // If the router is server-side rendered, we need to set the right response code and headers.
      // Without this, it's possible CDNs might cache the invalid path.
```

website/src/client/components/Router.tsx
```diff
@@ -6,42 +7,60 @@ import App from './App';
...
+      props.ctx.response.status = 404;
+      props.ctx.response.set(
+        'Cache-Control',
+        'private, no-cache, no-store, max-age=0, must-revalidate'
+      );
+    }
+
+    return <NonExistent />;
+  }, []);
```
ide, line 31:
```suggestion
  }, [props.ctx]);
```

website/src/server/index.tsx
```diff
@@ -74,6 +74,8 @@ if (process.env.COMPRESS_ASSETS === 'true') {
 app.use(sw());
 
 if (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'test') {
+  // Always try to send static files asap. For files that doesnt exist,
+  // it will server-side render the client router and set the appropriate status code.
```
ide, line 78:
```suggestion
  // Always try to send static files ASAP. For files that don't exist,
  // the server will server-side render the client router and set the appropriate status code.
```

website/src/client/components/Router.tsx
```diff
@@ -6,42 +7,60 @@ import App from './App';
...
-        <Route exact path="/@:username/:projectName+" render={this._renderRoute} />
-        <Route exact path="/:id" render={this._renderRoute} />
-        <Route exact path="/" render={this._renderRoute} />
-        <Route component={NonExistent} />
-      </Switch>
-    );
-  }
+    },
+    [render404]
+  );
```
ide, line 53: List out the props instead of using `...rest` and specify them as dependencies for the callback hook: data, queryParams, userAgent. This also makes sure ctx doesn't get passed into the components.

website/src/server/index.tsx
```diff
@@ -74,6 +74,8 @@ if (process.env.COMPRESS_ASSETS === 'true') {
 app.use(sw());
 
 if (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'test') {
+  // Always try to send static files ASAP. For files that don't exist,
+  // the server will server-side render the client router and set the appropriate status code.
   app.use(mount('/dist', serve(path.join(__dirname, '..', '..', 'dist'))));
```
byCedric, line 79: Add `no-cache` to all dist files, because of SW caching.

ide: `public, no-cache` is probably right for dist by default. You found an example of a dist file whose URL stays constant even when the file contents change.

Eventually for the dist files whose filenames are guaranteed to change whenever the file contents change, `public, max-age=31536000, immutable`, would be optimal.

website/src/client/components/Router.tsx
ide: The old `renderRoute` would always read from the latest props (`this.props`).

This new code is buggy if the Router component is re-rendered: since `render404` is the only Hook dependency, if it doesn't change, then we'll use the old, memoized value of `renderRoute` from a previous render, which references the old `props` variable.

To match the behavior of the code for this PR, there needs to be no dependency array: always redefine `renderRoute` so that it always references the latest `props` variable.

byCedric: Yep, I agree on the rerender part. But in SSR we don't rerender, and in the client, all the `rest` props from `props` are only used in `App` when actually initially mounting the component. Switching from embed to app (or reversed) isn't possible/irrelavant client-side, error pages (e.g. snack data has an error) only contain an absolute link that will trigger a reload (I believe), and the router props arent part of the rendering stuff.

That being said, I think it's fine to just create a new renderRoute function for a rerender. We don't actually do that in the Snack app itself, we just use these routes to start the App in different ways.

byCedric: Done, also the only item in `rest` turned out to be `userAgent`, so I updated that. Also moved it to a function instead of callback to avoid possible (but unlikely) `data` content changes.

ide (approved):
Looks great!

ide:
BTW there are two issues to look into: Prettier/Lint is failing in CI, and the Snack website deployment was having ImagePullBackoff issues. However maybe just deploying a new commit will fix things since CI should upload a new image.

website/src/server/index.tsx
byCedric: How about this approach (currently implemented in this PR). There are some other assets that also never change names, some of which are handled through some exclusion rules in webpack (not sure why).

To have a predictive rule on determining if the asset can be cached or not:
- Add a `.cached.` marker in the exported filename (mostly for chuncks that includes `[hash]`)
- Webpack worker plugin adds a `.worker.js` at the end for some of those files, but also these include `[hash]` and thus can be cached.

This would be an opt-in and shouldn't cause issues if we need to add more processing exclusion rules, they should not include the `.cached.` file segement.

<details><summary>These are all assets of website</summary>

```
dist/assets/018c0c569c00c0e707dcc30d1f40b354.png
dist/assets/1f423990e6bfdae184f966c86dc22c36.png
dist/assets/27b3706e74e4ee9f45ea34967a2f3fe4.png
dist/assets/48505df7e1a1eb25606c4789c3a3e235.png
dist/assets/52001a8b742a278688f6b516f931d81b.png
dist/assets/54da9e6c3dac08e684c23ebb5a3d2248.png
dist/assets/58b405f0e890b2e0c10e2881e13ae570.png
dist/assets/6b544ca2f9529bc2b22aa6978ebd5862.svg
dist/assets/6dcf149da3a69d5ee3fe7fb6933f5fa6.png
dist/assets/7bc60d33ce267a372e7d7b9807495282.png
dist/assets/7f658e7f03b868a6e934914996462329.png
dist/assets/9682e34e1f058feab75a232e3b9f33c6.png
dist/assets/af05a3786092b27e789607af7f5ab349.png
dist/assets/bc351fd24f9bd32bc131f122d42c1a77.svg
dist/assets/c51d82a43adccd86d66477bfc28dd39b.png
dist/assets/d14a681a76bdb229e305aa965b54dd18.png
dist/assets/d93a5d445d7b7a8be13f38362a9dcf7e.png
dist/assets/e392635154a17bd759d3bbb77e72eada.png
dist/assets/e95fe5167b7dfc17a242d6a9f10a60c3.png
dist/assets/ea07de06422d3379a4f579dfc4a779b9.png
dist/assets/f95e873b5154c5128ccb7bf69715156d.svg
dist/vendors~Editor/MonacoEditor.css
dist/vendors~Editor/MonacoEditor.css.map
dist/0.090843be60561cfc4f38.chunk.cached.worker.js
dist/0.1162b3ef4bc1c36fe5d4.chunk.cached.worker.js
dist/0.22e9a9be1201b9bd2641.chunk.cached.worker.js
dist/0.46d31f457e19cad8d352.chunk.cached.worker.js
dist/0.4dd0eddec8614de30d8c.chunk.cached.js
dist/0.7f09aa858bd49cbd2735.chunk.cached.worker.js
dist/1.4dd0eddec8614de30d8c.chunk.cached.js
dist/10.4dd0eddec8614de30d8c.chunk.cached.js
dist/12.4dd0eddec8614de30d8c.chunk.cached.js
dist/13.4dd0eddec8614de30d8c.chunk.cached.js
dist/14.4dd0eddec8614de30d8c.chunk.cached.js
dist/16.4dd0eddec8614de30d8c.chunk.cached.js
dist/17.4dd0eddec8614de30d8c.chunk.cached.js
dist/18.4dd0eddec8614de30d8c.chunk.cached.js
dist/19.4dd0eddec8614de30d8c.chunk.cached.js
dist/2.4dd0eddec8614de30d8c.chunk.cached.js
dist/20.4dd0eddec8614de30d8c.chunk.cached.js
dist/21.4dd0eddec8614de30d8c.chunk.cached.js
dist/22.4dd0eddec8614de30d8c.chunk.cached.js
dist/23.4dd0eddec8614de30d8c.chunk.cached.js
dist/24.4dd0eddec8614de30d8c.chunk.cached.js
dist/25.4dd0eddec8614de30d8c.chunk.cached.js
dist/26.4dd0eddec8614de30d8c.chunk.cached.js
dist/27.4dd0eddec8614de30d8c.chunk.cached.js
dist/28.4dd0eddec8614de30d8c.chunk.cached.js
dist/29.4dd0eddec8614de30d8c.chunk.cached.js
dist/3.4dd0eddec8614de30d8c.chunk.cached.js
dist/30.4dd0eddec8614de30d8c.chunk.cached.js
dist/31.4dd0eddec8614de30d8c.chunk.cached.js
dist/32.4dd0eddec8614de30d8c.chunk.cached.js
dist/33.4dd0eddec8614de30d8c.chunk.cached.js
dist/34.4dd0eddec8614de30d8c.chunk.cached.js
dist/35.4dd0eddec8614de30d8c.chunk.cached.js
dist/36.4dd0eddec8614de30d8c.chunk.cached.js
dist/37.4dd0eddec8614de30d8c.chunk.cached.js
dist/38.4dd0eddec8614de30d8c.chunk.cached.js
dist/39.4dd0eddec8614de30d8c.chunk.cached.js
dist/4.4dd0eddec8614de30d8c.chunk.cached.js
dist/40.4dd0eddec8614de30d8c.chunk.cached.js
dist/41.4dd0eddec8614de30d8c.chunk.cached.js
dist/42.4dd0eddec8614de30d8c.chunk.cached.js
dist/43.4dd0eddec8614de30d8c.chunk.cached.js
dist/44.4dd0eddec8614de30d8c.chunk.cached.js
dist/45.4dd0eddec8614de30d8c.chunk.cached.js
dist/46.4dd0eddec8614de30d8c.chunk.cached.js
dist/47.4dd0eddec8614de30d8c.chunk.cached.js
dist/48.4dd0eddec8614de30d8c.chunk.cached.js
dist/49.4dd0eddec8614de30d8c.chunk.cached.js
dist/5.4dd0eddec8614de30d8c.chunk.cached.js
dist/50.4dd0eddec8614de30d8c.chunk.cached.js
dist/51.4dd0eddec8614de30d8c.chunk.cached.js
dist/52.4dd0eddec8614de30d8c.chunk.cached.js
dist/53.4dd0eddec8614de30d8c.chunk.cached.js
dist/54.4dd0eddec8614de30d8c.chunk.cached.js
dist/55.4dd0eddec8614de30d8c.chunk.cached.js
dist/56.4dd0eddec8614de30d8c.chunk.cached.js
dist/57.4dd0eddec8614de30d8c.chunk.cached.js
dist/58.4dd0eddec8614de30d8c.chunk.cached.js
dist/59.4dd0eddec8614de30d8c.chunk.cached.js
dist/6.4dd0eddec8614de30d8c.chunk.cached.js
dist/60.4dd0eddec8614de30d8c.chunk.cached.js
dist/61.4dd0eddec8614de30d8c.chunk.cached.js
dist/62.4dd0eddec8614de30d8c.chunk.cached.js
dist/63.4dd0eddec8614de30d8c.chunk.cached.js
dist/64.4dd0eddec8614de30d8c.chunk.cached.js
dist/65.4dd0eddec8614de30d8c.chunk.cached.js
dist/66.4dd0eddec8614de30d8c.chunk.cached.js
dist/67.4dd0eddec8614de30d8c.chunk.cached.js
dist/68.4dd0eddec8614de30d8c.chunk.cached.js
dist/69.4dd0eddec8614de30d8c.chunk.cached.js
dist/7.4dd0eddec8614de30d8c.chunk.cached.js
dist/70.4dd0eddec8614de30d8c.chunk.cached.js
dist/71.4dd0eddec8614de30d8c.chunk.cached.js
dist/72.4dd0eddec8614de30d8c.chunk.cached.js
dist/73.4dd0eddec8614de30d8c.chunk.cached.js
dist/74.4dd0eddec8614de30d8c.chunk.cached.js
dist/75.4dd0eddec8614de30d8c.chunk.cached.js
dist/76.4dd0eddec8614de30d8c.chunk.cached.js
dist/77.4dd0eddec8614de30d8c.chunk.cached.js
dist/78.4dd0eddec8614de30d8c.chunk.cached.js
dist/79.4dd0eddec8614de30d8c.chunk.cached.js
dist/8.4dd0eddec8614de30d8c.chunk.cached.js
dist/80.4dd0eddec8614de30d8c.chunk.cached.js
dist/81.4dd0eddec8614de30d8c.chunk.cached.js
dist/82.4dd0eddec8614de30d8c.chunk.cached.js
dist/83.4dd0eddec8614de30d8c.chunk.cached.js
dist/84.4dd0eddec8614de30d8c.chunk.cached.js
dist/85.4dd0eddec8614de30d8c.chunk.cached.js
dist/86.4dd0eddec8614de30d8c.chunk.cached.js
dist/9.4dd0eddec8614de30d8c.chunk.cached.js
dist/EditorView.css
dist/EditorView.css.map
dist/Markdown/MarkdownPreview.css
dist/Markdown/MarkdownPreview.css.map
dist/app.bundle.js
dist/build-stats.js
dist/sw.bundle.js
```

</detauls>

PR #298, opened 2022-05-24
[website] Update URL on Device Preview Toggle

danstepanov:
# Why

Snack URLs already support the query param `platform`; however, when sharing the URL for a Snack, it does not reflect the `platform` in the URL. The desired UX here should just be copying and pasting the URL in the browser.

# How

Use `window.history.pushState()` to update the current url based on the platform selected

# Test Plan

1. Run `yarn && yarn build` from the root
2. Run `yarn start` from the root
3. Within the Device Preview window, toggle the device ("My Device", "iOS", "Android", "Web") and observe the URL change in the browser search bar.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -118,6 +118,44 @@ The code for the client is located under `src/client/`. The webpack build create
 
 Scripts related to deployment, like the Dockerfile, are under `deploy`. Note: even though the scripts are under `deploy`, you must run them from the **root of the repository**; they are sensitive to `cwd`.
 
+## Query Parameters
+
+You can pass certain parameters in the Snack URL to customize the behavior of the Snack.
+
+Here is a summary of all the parameter options that you can use:
+
+> All query parameter values must be URL encoded
+
+| parameter | default | description | type |
+| -------- | ------- | ----------- | ------ |
+| **dependencies** | | Comma separated list of <name>@<version>. | [SnackDependency](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L64) |
+| **description** | | The description of your Snack. | `string` |
+| **files** | | Allows you to inline files to load through a query param, useful for doc maintainers who don't want to maintain twice the code (both markdown and snack, itself). [Read more about files](https://github.com/expo/snack/blob/main/docs/snack-sdk.md#files-dependencies-and-the-state). | [`SnackFiles`](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L40) |
+| **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this parameter.  | `string` |
+| **platform** | `web` | The platform on which your Snack should be run in the Device Preview window. | `android \| ios \| mydevice \| web` |
+| **preview** | `true` | Toggle to show the Device Preview. | `boolean` |
+| **sdkVersion** | _[default SDK](../packages/snack-content/src/defaults.ts#L3)_ | The Expo SDK version that your Snack should use. | [SDKVersion](../packages/snack-content/src/sdks/types.ts#L4) |
+| **sourceUrl** |  | One of two ways to send a file, via publicly-accessible URL of a JS file. | `string` |
+| **supportedPlatforms** | All platforms | Specify which platforms your Snack supports | `android \| ios \| mydevice \| web` |
+| **theme** | `light` | The visual theme of your Snack. | `light \| dark` |
+
+### Examples of Snack URLs
+
+All of these examples should be prefixed with `https://snack.expo.dev/`.
+
+| parameter | example |
+| --------- | ------- |
+| **name** | _[`?name=Gordon%20Freeman`](https://snack.expo.dev/?name=Gordon%20Freeman)_ |
+| **description** | _[`?description=Where%20is%20Half-Life%203`](https://snack.expo.dev/?description=Where%20is%20Half-Life%203)_ |
+| **platform** | _[`?platform=mydevice`](https://snack.expo.dev/?platform=mydevice)_ |
+| **preview** | _[`?preview=false`](https://snack.expo.dev/?preview=false)_ |
+| **sdkVerion** | _[`?sdkVersion=45.0.0`](https://snack.expo.dev/?sdkVersion=45.0.0)_ |
+| **supportedPlatforms** | _[`?supportedPlatforms=android,ios`](https://snack.expo.dev/?supportedPlatforms=android,ios)_ |
+| **theme** | _[`?theme=dark`](https://snack.expo.dev/?theme=dark)_ |
+| **dependencies** | _[`?dependencies=%40expo%2Fvector-icons%40*%2C%40react-native-community%2Fmasked-view`](https://snack.expo.dev/?dependencies=%40expo%2Fvector-icons%40*%2C%40react-native-community%2Fmasked-view)_ |
+| **files** | _[`?files=`](https://snack.expo.dev/?files={"type": "CODE", "contents": "alert('hello');" })_ |
+| **sourceUrl** | _[`?sourceUrl=https://reactnavigation.org/examples/6.x/hello-react-navigation.js`](https://snack.expo.dev/?sourceUrl=https://reactnavigation.org/examples/6.x/hello-react-navigation.js)_ |
+
 ## Running the tests
 
 We run unit tests with Jest. Run `yarn test` in another terminal to start Jest and have it continuously watch for changes. You also can run `yarn jest` if you want to run Jest without the watcher. Keep unit tests fast so that the feedback loop from them is fast.
--- a/website/README.md
+++ b/website/README.md
@@ -128,7 +128,7 @@ Here is a summary of all the parameter options that you can use:
 
 | parameter | default | description | type |
 | -------- | ------- | ----------- | ------ |
-| **dependencies** | | Comma separated list of <name>@<version>. | [SnackDependency](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L64) |
+| **dependencies** | | Comma separated list of \<name\>@\<version\>. | [SnackDependency](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L64) |
 | **description** | | The description of your Snack. | `string` |
 | **files** | | Allows you to inline files to load through a query param, useful for doc maintainers who don't want to maintain twice the code (both markdown and snack, itself). [Read more about files](https://github.com/expo/snack/blob/main/docs/snack-sdk.md#files-dependencies-and-the-state). | [`SnackFiles`](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L40) |
 | **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this parameter.  | `string` |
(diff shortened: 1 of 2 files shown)


website/README.md
```diff
@@ -118,6 +118,24 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| **platform** | `web` | The platform on which your Snack should be run in the Device Preview window. | `android`, `ios`, `mydevice`, `web` |
+| **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this prop.  | `string` |
+| **dependencies** |  | Dependencies to be used by your Snack beyond common dependencies and the ones added in the Snack's `package.json`. [Read more about Snack dependencies.]() | _`string[https://github.com/expo/snack/blob/main/docs/snack-sdk.md#using-dependencies]`_ |
+| **sdkVersion** | `44.0.0` | The Expo SDK version that your Snack should use. | `45.0.0`, `44.0.0`, `43.0.0`
+| **supportedPlatforms** | |  | `string[]` |
+| **files** | - |  |  |
+| **snackId** |  |  |  |
+| **description** |  | The description of your Snack | `string`
+| **preview** | `false` |  | `boolean`
+| **theme** | `light` | The visual theme of your Snack | `light`, `dark`
```
byCedric, line 137: 🙏

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| `name` | `?name=Gordon%20Freeman` |
+| `description` | `?description=Research%20assistant%20at%20the%20Black%20Mesa%20Research%20Facility.` |
+| `snackId` |  |
+| `platform` | `?platform=mydevice` |
+| `preview` | `?preview=false` |
+| `sdkVerion` | `?sdkVersion=45.0.0` |
+| `supportedPlatforms` | `?supportedPlatforms=android,ios` |
+| `theme` | `?theme=dark` |
+| `dependencies` | `?dependencies=%40expo%2Fvector-icons%40*%2C%40react-native-community%2Fmasked-view` |
+| `files` | `?files=` |
```
byCedric, line 155: For these examples, I think we can make it a bit shorter. What do you think of listing a couple, e.g. 3, URLs and explaining each param set? E.g. something like:

---

### Example URLs

- [?name=Gordon%20Freeman&description=Where%20is%20Halflife%203](https://snack.expo.dev/?name=Gordon%20Freeman&description=Where%20is%20Halflife%203)
  `name` is set to **Gordon Freeman**
  `description` is set to **Where is Halflife 3**

- [?preview=false&theme=dark](https://snack.expo.dev/?preview=false&theme=dark)
  `preview` is **false**, hiding the right "preview" panel
  `theme` is **dark**

---

I'm also ok with a table, but let's try and make that a bit cleaner.
- Try and use **bold** for query param names, I think it reads a bit nicer plus they should be double-click selectable as they are 
URL safe (no spaces). You already did this nicely on the query param table
- Make the example links clickable (just like I did in the example above).

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| **files** | | Allows you to inline files to load through a query param, useful for doc maintainers who don't want to maintain twice the code (both markdown and snack, itself). [Read more about files](https://github.com/expo/snack/blob/main/docs/snack-sdk.md#files-dependencies-and-the-state). | [`SnackFiles`](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L40) |
+| **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this parameter.  | `string` |
+| **platform** | `web` | The platform on which your Snack should be run in the Device Preview window. | `android`, `ios`, `mydevice`, `web` |
+| **preview** | `true` | Toggle to show the Device Preview. | `boolean` |
+| **sdkVersion** | `44.0.0` | The Expo SDK version that your Snack should use. | `45.0.0`, `44.0.0`, `43.0.0` |
+| **snackId** | | The unique ID of your Snack. | `string` |
+| **supportedPlatforms** | All platforms | Specify which platforms your Snack supports | `android`, `ios`, `mydevice`, `web` |
+| **theme** | `light` | The visual theme of your Snack. | `light`, `dark` |
+
+## Examples of Snack URLs
```
byCedric, line 140:
```suggestion
### Examples of Snack URLs
```

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| **description** | | The description of your Snack. | `string` |
+| **files** | | Allows you to inline files to load through a query param, useful for doc maintainers who don't want to maintain twice the code (both markdown and snack, itself). [Read more about files](https://github.com/expo/snack/blob/main/docs/snack-sdk.md#files-dependencies-and-the-state). | [`SnackFiles`](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L40) |
+| **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this parameter.  | `string` |
+| **platform** | `web` | The platform on which your Snack should be run in the Device Preview window. | `android`, `ios`, `mydevice`, `web` |
+| **preview** | `true` | Toggle to show the Device Preview. | `boolean` |
+| **sdkVersion** | `44.0.0` | The Expo SDK version that your Snack should use. | `45.0.0`, `44.0.0`, `43.0.0` |
+| **snackId** | | The unique ID of your Snack. | `string` |
```
byCedric, line 136: I think the **snackId** relates to either a draft snack ID or the snack channel ID/Session. In both cases, I don't think users should be interacting with that. Or is it relating to something else that's important for our users?

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| **supportedPlatforms** | All platforms | Specify which platforms your Snack supports | `android`, `ios`, `mydevice`, `web` |
+| **theme** | `light` | The visual theme of your Snack. | `light`, `dark` |
```
byCedric, line 138: What do you think of defining the available values more like typescript? e.g. 
```suggestion
| **theme** | `light` | The visual theme of your Snack. | `light | dark` |
```

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| **description** | | The description of your Snack. | `string` |
+| **files** | | Allows you to inline files to load through a query param, useful for doc maintainers who don't want to maintain twice the code (both markdown and snack, itself). [Read more about files](https://github.com/expo/snack/blob/main/docs/snack-sdk.md#files-dependencies-and-the-state). | [`SnackFiles`](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L40) |
+| **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this parameter.  | `string` |
+| **platform** | `web` | The platform on which your Snack should be run in the Device Preview window. | `android`, `ios`, `mydevice`, `web` |
+| **preview** | `true` | Toggle to show the Device Preview. | `boolean` |
+| **sdkVersion** | `44.0.0` | The Expo SDK version that your Snack should use. | `45.0.0`, `44.0.0`, `43.0.0` |
```
byCedric, line 135: We likely will forget to update this, so I think it would be better to just link the types 😅 

```suggestion
| **sdkVersion** | _[default SDK](../packages/snack-content/src/defaults.ts#L3)_ | The Expo SDK version that your Snack should use. | [SDKVersion](../packages/snack-content/src/sdks/types.ts#L4) |
```

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| -------- | ------- | ----------- | ------ |
+| **dependencies** | | Dependencies to be used by your Snack beyond common dependencies and the ones added in the Snack's `package.json`. [Read more about Snack dependencies](_`string[https://github.com/expo/snack/blob/main/docs/snack-sdk.md#using-dependencies]`_). | [SnackDependency](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L64) |
+| **description** | | The description of your Snack. | `string` |
+| **files** | | Allows you to inline files to load through a query param, useful for doc maintainers who don't want to maintain twice the code (both markdown and snack, itself). [Read more about files](https://github.com/expo/snack/blob/main/docs/snack-sdk.md#files-dependencies-and-the-state). | [`SnackFiles`](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L40) |
+| **name** | | The name of your Snack. When creating a new Snack, a random name is assigned unless one is provided via this parameter.  | `string` |
+| **platform** | `web` | The platform on which your Snack should be run in the Device Preview window. | `android`, `ios`, `mydevice`, `web` |
```
byCedric, line 133: Also here, I think using `android | ios | web | mydevice` would be nice?

byCedric (approved):
Looking good, let's fill in the rest and then merge it

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
...
 
+## Query Parameters
+
+You can pass certain parameters in the Snack URL to customize the behavior of the Snack.
+
+Here is a summary of all the parameter options that you can use:
+
+| parameter | default | description | type |
+| -------- | ------- | ----------- | ------ |
+| **dependencies** | | Dependencies to be used by your Snack beyond common dependencies and the ones added in the Snack's `package.json`. [Read more about Snack dependencies](_`string[https://github.com/expo/snack/blob/main/docs/snack-sdk.md#using-dependencies]`_). | [SnackDependency](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L64) |
```
byCedric, line 129: For the dependencies, I think it would be easier to just mention:
> Comma separated list of `<name>@<version>`

That should be accurate of what we expect here instead of a URL encoded object.

website/README.md
```diff
@@ -118,6 +118,42 @@ The code for the client is located under `src/client/`. The webpack build create
 
 Scripts related to deployment, like the Dockerfile, are under `deploy`. Note: even though the scripts are under `deploy`, you must run them from the **root of the repository**; they are sensitive to `cwd`.
 
+## Query Parameters
+
+You can pass certain parameters in the Snack URL to customize the behavior of the Snack.
+
+Here is a summary of all the parameter options that you can use:
+
+| parameter | default | description | type |
```
byCedric, line 127: Not sure how to emphasize this best, but I think it would be good to do so.

```suggestion
## Query Parameters

You can pass certain parameters in the Snack URL to customize the behavior of the Snack.

Here is a summary of all the parameter options that you can use:

> All query parameter values must be URL encoded

| parameter | default | description | type |
```

website/README.md
danstepanov: I'll clean up the table for now, I think it's important to show examples for `files` and `dependencies`, in particular. I'd be down to do 3 examples if they included those two. Lmk what you think about adding those (seems like linking `files` might not work unless we set up a hosted file).

website/README.md
danstepanov: it's mentioned in [SnackEmbed](https://github.com/expo/expo/blob/428b6b44d8c60a95f297b6623d98ebdfd4d5623e/docs/components/plugins/SnackEmbed.tsx#L6)

website/README.md
```diff
@@ -118,6 +118,44 @@ The code for the client is located under `src/client/`. The webpack build create
...
+| --------- | ------- |
+| **name** | _[`?name=Gordon%20Freeman`](https://snack.expo.dev/?name=Gordon%20Freeman)_ |
+| **description** | _[`?description=Where%20is%20Half-Life%203`](https://snack.expo.dev/?description=Where%20is%20Half-Life%203)_ |
+| **platform** | _[`?platform=mydevice`](https://snack.expo.dev/?platform=mydevice)_ |
+| **preview** | _[`?preview=false`](https://snack.expo.dev/?preview=false)_ |
+| **sdkVerion** | _[`?sdkVersion=45.0.0`](https://snack.expo.dev/?sdkVersion=45.0.0)_ |
+| **supportedPlatforms** | _[`?supportedPlatforms=android,ios`](https://snack.expo.dev/?supportedPlatforms=android,ios)_ |
+| **theme** | _[`?theme=dark`](https://snack.expo.dev/?theme=dark)_ |
+| **dependencies** | _[`?dependencies=%40expo%2Fvector-icons%40*%2C%40react-native-community%2Fmasked-view`](https://snack.expo.dev/?dependencies=%40expo%2Fvector-icons%40*%2C%40react-native-community%2Fmasked-view)_ |
+| **files** | _[`?files=`](https://snack.expo.dev/?files={"type": "CODE", "contents": "alert('hello');" })_ |
```
byCedric, line 156: The files params needs to be url encoded in the link, it breaks the markdown right now

Screenshot 2022-05-27 at 15 50 41

website/README.md
```diff
@@ -118,6 +118,44 @@ The code for the client is located under `src/client/`. The webpack build create
...
+
+You can pass certain parameters in the Snack URL to customize the behavior of the Snack.
+
+Here is a summary of all the parameter options that you can use:
+
+> All query parameter values must be URL encoded
+
+| parameter | default | description | type |
+| -------- | ------- | ----------- | ------ |
+| **dependencies** | | Comma separated list of <name>@<version>. | [SnackDependency](https://github.com/expo/snack/blob/main/packages/snack-content/src/types.ts#L64) |
```
byCedric, line 131: I think we need to escape the `<` and `>` by adding a backslash to it, e.g. `\<`. Right now, github filters it because it thinks its a HTML tag

Screenshot 2022-05-27 at 15 51 06

byCedric (approved):
Left 2 comments to fix something broken, after that feel free to merge 😄  Thanks for this! 🙏

website/README.md
byCedric: Lets leave it then :)

website/README.md
danstepanov: good catch

website/README.md
danstepanov: made the change, just as a heads up, the provided url does not work when you click through, not entirely sure if this is to be expected but I think it's a bad experience

danstepanov: the error shown is `Invalid files no entry point found`

danstepanov:
going to merge for now and can open a separate PR to fix the files link, assuming that the expected experience is different

Issue #299, opened 2022-05-24
The runtime does not package the babel helpers while snackager uses runtime-transform


mmomtchev:
### Summary

My module exports a `main` and a `module` with the `module` using ES6 syntax which makes Babel to insert a couple of helpers:

```shell
yarn bundle @mmomtchev/react-native-settings@1.0.3 | ./node_modules/.bin/bunyan
```
```js
file: /tmp/@mmomtchev/react-native-settings/web-bundle.js
--
externals: [
  "@babel/runtime/helpers/interopRequireDefault",
  "@babel/runtime/helpers/slicedToArray",
  "@react-navigation/native-stack",
  "react",
  "react-native"
]
```

### What platform(s) does this occur on?

Web

### SDK Version

v44.0.0

### Reproducible demo or steps to reproduce from a blank project

https://snack.expo.dev/@mmomtchev/@mmomtchev-react-native-settings

PR #302, opened 2022-05-25
[snackager] correct output with package type: module

mmomtchev:
# Why

Refs: https://github.com/expo/snack/issues/299
Refs: https://github.com/microsoft/rnx-kit/issues/142

The Metro Babel preset does not produce correct output
for packages containing "type": "module" unless
the `disableImportExportTransform` option is used to
prevent it from including @babel/plugin-transform-modules-commonjs

# How

I disable `disableImportExportTransform` option if I detect `"type": "module"`

# Test Plan

I included my module in the test suite


byCedric:
Awesome, also confirmed your fix is working as intended for your package, so disregard the CI :). I'll try and find other packages with similar `type: module` settings to check for unintended side effects. Probably this weekend!

PR #307, opened 2022-06-13
[website] remove legacy workaround for SDK 39 and below

Simek:
# Why

We have dropped support for SDK 39 a while ago.

# How

Remove the legacy web preview workaround.

# Test Plan

The web preview was loading correctly when running Snack stack locally.


byCedric (approved):
LGTM

PR #311, opened 2022-07-19
[snackager] Add missing externals for @sentry/react-native

byCedric:
# Why

`@sentry/react-native@3.4.2` contains some imports to React Native internals, we need to externalize that to make it bundle in snackager.

# How

- Added snackager externals
- Added runtime aliases

# Test Plan

See if CI passes and staging can bundle `@sentry/react-native@3.4.2`


byCedric:
Bundle test: https://github.com/expo/snack/runs/7411287031?check_suite_focus=true

PR #315, opened 2022-07-20
Upgrade Snack to Expo SDK 46 beta

byCedric:
# Why

We need to upgrade to SDK 46, this is a test run to see what will happen during the beta.

# How

Followed the steps in [./docs/expo-sdk-upgrade.md](https://github.com/expo/snack/blob/main/docs/expo-sdk-upgrade.md)

# Test Plan

After everything seems ok, ~~I'll manually deploy it to staging (without merging to main).~~ Done!


byCedric:
Runtime deployment issues were caused by the new CLI. I replaced `expo build:web` with `expo export:web`, and force-used the old Expo CLI for `expo-cli publish`.

runtime/src/NativeModules/Amplitude.native.tsx
```diff
@@ -1,3 +1,11 @@
-import * as Amplitude from 'expo-analytics-amplitude';
+// Disable Amplitude for SDK 46
```
brentvatne, line 1: ah - i didn't realize that we were using this in the snack runtime 😰

brentvatne (commented):
i'm not too familiar with snack but this all looks pretty reasonable. looking forward to seeing the solution for skia support :) let me know if you need anything from me

runtime/src/NativeModules/Amplitude.native.tsx
byCedric: I already pinged @tcdavis, but we aren't using analytics from the Runtime anymore. So should be good :)

byCedric:
Change of plans, I'm squashing all the SDK 46 PRs into `sdk/46`, so we can merge it in main once the stable SDK is out cc @brentvatne. Helps me validate a few things, but still be able to update runtime to whatever we need for the stable release.

Issue #318, opened 2022-07-21
Snack doesn't support `await import()` statements


byCedric:
### Summary

While React Suspense works pretty nicely, the lazy imports doesn't seem to be connected to the Snack files. Instead, it tries to load it inside the published Runtime files (e.g. on web, it would treat it as asset).

### What platform(s) does this occur on?

Web

### SDK Version

46

### Reproducible demo or steps to reproduce from a blank project

TBD

byCedric:
Fixed in #333

PR #325, opened 2022-07-21
[runtime] Add Skia support for web

byCedric:
# Why

On Android and iOS, Skia seems to work nicely already. But for web, we need to initialize `canvaskit-wasm`. This adds that support.

# How

- Added Snack-specific initialization in the Runtime, [based on the original web init](https://github.com/Shopify/react-native-skia/blob/5adbade804191deef4380ae42796d3b5f050dad7/package/src/web/WithSkia.tsx)
- Note, we have to use `require` instead of `await import` due to #318  

# Test Plan

~~Will share a snack for staging.~~

Here it is! This should work on Android, iOS, and web: https://staging.snack.expo.dev/@bycedrictest/universal-skia-demo.

> Note
> The Android and iOS tabs are linked to Appetize, which isn't updated yet. Use Expo Go `>=2.25.0`.


byCedric:
cc @wcandillon

runtime/src/NativeModules/CanvasKit.web.tsx
```diff
@@ -0,0 +1,21 @@
+// Types are monkey-patched in runtime/types/canvaskit-wasm.d.ts
+import type { CanvasKit as CanvasKitType } from 'canvaskit-wasm';
+import CanvasKitInit from 'canvaskit-wasm/bin/full/canvaskit';
+import { version } from 'canvaskit-wasm/package.json';
```
wcandillon, line 4: Haha this is such a cool trick 😅

brentvatne:
is this ready to move to ready for review?

byCedric:
@brentvatne It's reviewable, but I'd need to update the other SDK 46 beta upgrade to non-beta when it's out 😄 So I'd prefer to keep this separate until the SDK 46 upgrade is complete.

PR #329, opened 2022-07-25
[snack-babel-standalone] Add eslint bundle and documentation

byCedric:
# Why

The current vendored ESLint system is super hard to maintain or upgrade. This change allows us to simplify that process by creating a similar `snack-eslint-standalone` package, with the `@babel/eslint-parser` bundled and "hot-wired" to `snack-babel-standalone/eslint`.

# How

- Added specific entry point for ESLint
- Tested building a bundle for `snack-eslint-standalone` with `@babel/core` aliased to `snack-babel-standalone@^2.0.0`.

# Test Plan

This change will be added to the SDK 46 upgrade and should be available on Staging soon.

> Note that this PR doesn't affect anything on its own, as this package is decoupled from the monorepo system.


byCedric:
Something we could do is add a bundle-size check to the `snack-babel-standalone` package and `snack-eslint-standalone`. That should keep the bundle size visible, as the previous one grew to 28mb.

byCedric:
@brentvatne going to merge this now, bc I need it for staging.

PR #332, opened 2022-07-26
[snack-sdk] Upgrade pubnub to 7.2.0

byCedric:
# Why

This drops the dependency chain including unimodules, and makes pubnub so much more light-weight.

# How

- Upgraded pubnub to 7.2.0
- Added UUID generation per transport instance

# Test Plan

Probably have to check with a partner on this one, and probably good to bump up with a breaking version?


yarn.lock
```diff
@@ -3668,21 +3635,6 @@
     "@typescript-eslint/types" "5.18.0"
     eslint-visitor-keys "^3.0.0"
 
-"@unimodules/core@*":
```
byCedric, line 3671: BE GONE

byCedric:
TIL: jest wasn't configured right all these years

PR #361, opened 2022-11-23
Fix initializing embedded snacks using `data-snack-files`

NickGerleman:
# Why

`data-snack-files` allows specifying specific filenames for code, instead of `data-snack-code` which is placed as `App.js`, or `data-snack-id` which points to a saved Snack. `data-snack-files` replaces `data-snack-code` if present, but an element with only the former will not be initialized.

# How

This adds `data-snack-files` to the list of searched for elements, to catch embedded snacks not using `data-snack-code` (e.g. to embed TypeScript instead of JavaScript).

# Test Plan

Before the change, using data-snack-files in the react-native website would lead to blank snack examples when switching through tabs. Snacks show correctly after monkey-patching in the new function.

### DIFF
--- a/website/src/server/EmbeddedSnackScript.tsx
+++ b/website/src/server/EmbeddedSnackScript.tsx
@@ -52,7 +52,7 @@ export const script = `
       var iframe = document.createElement('iframe');
 
       var iframeQueryParams = '?iframeId=' + iframeId;
-      
+
       if (options.preview) {
         iframeQueryParams += '&preview=' + options.preview;
       }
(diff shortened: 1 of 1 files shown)


NickGerleman:
A bit orthogonal, I wonder if `data-snack-code` should just default to `tsx` instead of `js`, given one is mostly the superset of the other (ignoring Flow). The issue in this PR came about in making that change in the RN website.

Simek (approved):
Thank you for catching that, LGTM! 👍

byCedric:
Thanks @NickGerleman! I'll merge and deploy this tomorrow :)

byCedric:
Alright, it's going to production now!

PR #365, opened 2022-12-06
[snack-sdk][website] Fix local webplayer hanging in connecting state

Kudo:
# Why

local webplayer is hanging in connecting state.

# How

the `Snack` sdk class only supports fixed webplayer url in construction. after switching to local webplayer, the transport still send message to old origin (our cloud `SNACK_WEBPLAYER_URL`). that's why website cannot connect to runtime correctly.

this pr adds a `setWebPlayerURL` function to change the webplayer URL in snack-sdk. once we changed the url, we should also re-create a new webplayer transport.

# Test Plan

# unit test

```
 PASS  src/__tests__/webplayer-test.ts
  webpreview
    ✓ updates a webplayer url (1 ms)
```

# integration test

1. `yarn start` in workspace root
2. `NODE_OPTIONS='--openssl-legacy-provider' npx expo start -w` in runtime
3. open http://snack.expo.test and switch sdk to localhost

### DIFF
--- a/packages/snack-sdk/src/Session.ts
+++ b/packages/snack-sdk/src/Session.ts
@@ -88,6 +88,7 @@ export default class Snack {
   private state: SnackState;
   private stateListeners: Set<SnackStateListener> = new Set();
   private logListeners: Set<SnackLogListener> = new Set();
+  private readonly options: SnackOptions;
   private readonly createTransport: (options: SnackTransportOptions) => SnackTransport;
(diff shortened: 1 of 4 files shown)


byCedric (approved):
Thanks, looks good, I haven't tested it, though, but we could send it to staging after merging.

Kudo:
thanks Cedric! let me try my first merge for snack.

PR #366, opened 2022-12-06
Fix syntax highlighting for embedded tsx files

NickGerleman:
Syntax highlighting for embedded snacks using tsx is poor in comparison to jsx. The issue does not show in the full website.

**jsx**
image

**tsx**
image

I haven't tried to test this, but looking through to find where syntax highlighting was happening, I noticed that `SimpleEditor` was setting a language of 'typescript' for tsx files, instead of 'tsx' https://github.com/PrismJS/prism/blob/master/components/prism-tsx.js This seems suspiciously like the issue being seen, so this change makes it so that tsx files in SimpleEditor set the correct language.

I am not familiar with this code, so please let me know if this isn't the right place for embedded snacks, or if there is a way to easily test.

### DIFF
--- a/website/src/client/components/Editor/SimpleEditor.tsx
+++ b/website/src/client/components/Editor/SimpleEditor.tsx
@@ -12,6 +12,7 @@ import 'prismjs/components/prism-markup';
 import 'prismjs/components/prism-jsx';
 import 'prismjs/components/prism-json';
 import 'prismjs/components/prism-markdown';
+import 'prismjs/components/prism-tsx';
 import withThemeName, { ThemeName } from '../Preferences/withThemeName';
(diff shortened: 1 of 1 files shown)


NickGerleman:
It looks like the build workflow might be having an unrelated issue 
image

byCedric:
Yeah, the secrets are not available in a fork. I'll copy this branch to the original repo and we can see if that works or not ;)

byCedric (approved):
gentleman-club

byCedric:
@NickGerleman, I also fixed our CI to not-fail when opening a PR from a fork. 😄 (see [this pr](https://github.com/expo/snack/pull/369)). Next time, it just gets validated properly

PR #367, opened 2022-12-07
Fix syntax highlighting for embedded tsx files

byCedric:
This is a copy of #366 that we can validate in CI and push to staging. DO NOT MERGE THIS PR.

Syntax highlighting for embedded snacks using tsx is poor in comparison to jsx.

I haven't tried to test this, but looking through to find where syntax highlighting was happening, I noticed that `SimpleEditor` was setting a language of 'typescript' for tsx files, instead of 'tsx' https://github.com/PrismJS/prism/blob/master/components/prism-tsx.js This makes it so that tsx files in SimpleEditor set the correct language.

I am not familiar with this code, so please let me know if this isn't the right place for embedded snacks, or if there is a way to easily test.

### DIFF
--- a/website/src/client/components/Editor/SimpleEditor.tsx
+++ b/website/src/client/components/Editor/SimpleEditor.tsx
@@ -12,6 +12,7 @@ import 'prismjs/components/prism-markup';
 import 'prismjs/components/prism-jsx';
 import 'prismjs/components/prism-json';
 import 'prismjs/components/prism-markdown';
+import 'prismjs/components/prism-tsx';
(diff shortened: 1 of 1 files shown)


NickGerleman:
I connected the RN website to https://staging.snack.expo.dev/embed.js to test, and verified we are making network requests to https://staging.snack.expo.dev, but am not seeing a change in highlighting.

Looking at the minifed bundle being served, I'm not seeing the change though. 

image

image

NickGerleman:
This is confirmed to fix highlighting in the current version served from production 👍.

PR #372, opened 2022-12-09
[website] Remove device ID option for linking Snack to Expo Go

brentvatne:
# Why

It's not working currently, and this feature is unnecessary / not worth fixing. Account or QR code seem sufficient.

Related
- https://github.com/expo/expo/pull/19939
- https://linear.app/expo/issue/ENG-6889/device-id-linking-in-snack-appears-to-be-broken

# How

Remove all code related to device ID.

# Test Plan

- [ ] Run the app locally in embedded mode, make sure it works and UI looks fine. Any tips on how to do this @byCedric?
- [x] Run the app locally, ensure that device ID tab isn't in the "Run on device" modal.

image


byCedric:
Merged through #472

PR #373, opened 2022-12-21
Prepare for Google front-end servers

nicknovitski:
# Why

There's some things that are beneficial or required before I can set up our new front end servers to handle requests to the snack website and snackager.

# How

- [website] Don't use stoppable
- Set keep-alive timeout to 10m20s
- Connect k8s services to Google Backend Services
- [website] Fail readiness probe when shutting down

### DIFF
--- a/website/src/server/index.tsx
+++ b/website/src/server/index.tsx
@@ -143,6 +143,19 @@ httpServer.keepAliveTimeout =
 // Listen to HTTP server error events and handle shutting down the server gracefully
 let exitSignal: ShutdownSignal | null = null;
 let httpServerError: Error | null = null;
+let ready = true;
+
+app.use(
+  mount('/ready', (ctx) => {
+    if (ready) {
+      ctx.response.status = 200;
+      ctx.body = 'ready';
+    } else {
+      ctx.response.status = 503;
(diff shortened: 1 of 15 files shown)


website/src/server/index.tsx
```diff
@@ -136,24 +135,38 @@ const httpServer = app.listen(port, host, backlog, () => {
...
+  if (ready) {
+    ctx.response.status = 200;
+    ctx.body = 'ready';
+  } else {
+    ctx.response.status = 503;
+    ctx.body = 'shutting down';
+  }
+});
+
+app.use(mount('/ready', readiness));
```
ide, line 160: koa-mount also supports mounting middleware directly so there's no need to create a whole Koa object.

ide, line 160: koa-mount also supports mounting middleware directly so there's no need to create a whole Koa object.
```suggestion
app.use(mount('/ready', function readiness(ctx) {
  if (ready) {
    ctx.response.status = 200;
    ctx.body = 'ready';
  } else {
    ctx.response.status = 503;
    ctx.body = 'shutting down';
  }
}));
```

ide (approved):

randomhajile (approved):

PR #376, opened 2023-01-03
[snack-sdk] add snackpub integration

Kudo:
# Why

close ENG-7101

## pr series

👉 #376
#377
#378 
#379
#380
#381 

# How

- clean webplayer transport code that it still relied on pubnub transport
- refactor transports so now we have three different transports
  - TransportImplPubNub
  - TransportImplWebPlayer
  - TrafficMirroringTransport = TransportImplPubNub + TransportImplSocketIO with traffic mirror
- original `createTransport` will create `TransportImplPubNub` and new `createSnackPubTransport` will create `TrafficMirroringTransport`

# Test Plan

- snackpub integration tests
- unit tests passed (refactor from PubNub mock to TestTransport mock)


packages/snack-sdk/src/transports/CompositedTransport.ts
```diff
@@ -0,0 +1,34 @@
+import TransportImplPubNub from './TransportImplPubNub';
+import TransportImplSocketIO from './TransportImplSocketIO';
+import {
+  SnackTransport,
+  SnackTransportListener,
+  SnackTransportMessage,
+  SnackTransportOptions,
+} from './types';
+
+export default class CompositedTransport implements SnackTransport {
```
byCedric, line 10: Generally speaking, I prefer to use the present tense in code (unless the context requires another tense version). It's a personal preference, but maybe worth to do :)

```suggestion
export default class CompositeTransport implements SnackTransport {
```

> **Note**
> This change would also require the file name to change I think

packages/snack-sdk/src/transports/TransportImplPubNub.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+  };
+
+  private onPubNubMessage = (event: PubNub.MessageEvent) => {
+    const message: ProtocolIncomingMessage = event.message;
+    this.handleMessage(event.publisher, message);
+  };
+
+  private onPubNubStatus = (_event: PubNub.StatusEvent) => {
+    // TODO: Do something here?
```
byCedric, line 119: haha, nice, do we need to do anything here? Maybe some things that might be worth ([using this doc](https://www.pubnub.com/docs/sdks/javascript/status-events#subscription))

- Disconnect/disable transport when the connection was lost (or even try to reconnect)
- Maybe try to clean up the pubnub cache when running into `PNRequestMessageCountExceededCategory`

packages/snack-sdk/src/transports/TransportImplSocketIO.ts
```diff
@@ -0,0 +1,104 @@
...
+
+interface ClientToServerEvents {
+  message: (data: { channel: string; message: ProtocolOutgoingMessage; sender: string }) => void;
+  subscribeChannel: (data: { channel: string; sender: string }) => void;
+  unsubscribeChannel: (data: { channel: string; sender: string }) => void;
+}
+
+export default class TransportImplSocketIO extends TransportImplBase {
+  private readonly _snackpubURL: string;
+  private _socket: Socket<ServerToClientEvents, ClientToServerEvents> | null = null;
```
byCedric, line 22: Another personal preference, feel free to ignore. But with TypeScript, we can already define `private`/`public`/`protected`, we don't need to do that in the name anymore :)

```suggestion
  private readonly snackpubURL: string;
  private socket: Socket<ServerToClientEvents, ClientToServerEvents> | null = null;
```

packages/snack-sdk/src/transports/TransportImplWebPlayer.ts
```diff
@@ -0,0 +1,79 @@
...
+  webPlayerURL: string;
+  window: Window;
+  ref: SnackWindowRef;
+}
+
+export default class TransportImplWebPlayer extends TransportImplBase {
+  private _currentWindow: Window | undefined;
+  private _targetWindowRef: SnackWindowRef | undefined;
+  private _isStarted: boolean;
+  private readonly _origin: string;
```
byCedric, line 19: Same here :)

> Another personal preference, feel free to ignore. But with TypeScript, we can already define private/public/protected, we don't need to do that in the name anymore :)

```suggestion
  private currentWindow: Window | undefined;
  private targetWindowRef: SnackWindowRef | undefined;
  private isStarted: boolean;
  private readonly origin: string;
```

byCedric: Also, random sanity check, is `started` the only status this transport can have? If so, ignore this suggestion, if there are more, maybe this would be more appropriate.

```suggestion
  private _currentWindow: Window | undefined;
  private _targetWindowRef: SnackWindowRef | undefined;
  private status: 'stopped' | 'started' = 'stopped';
  private readonly _origin: string;
```

packages/snack-sdk/src/transports/types.ts
```diff
@@ -45,9 +45,7 @@ export type SnackTransportEventConnect = {
 export type SnackTransportEventDisconnect = {
   type: 'disconnect';
   connectionId: string;
-  data: {
-    timedout: boolean;
-  };
+  data: any;
```
byCedric, line 48: Do we need this `data` property on the disconnect event? If not, I would omit it. The type doesn't really add much value if we don't use it. (and if we do use it, might be worth adding proper types)

byCedric (commented):
Looks great, had some minor comments about some small things. Thanks for adding tests :)

packages/snack-sdk/src/transports/types.ts
Kudo: this is required because `SnackTransportEventDisconnect` is one kind of `SnackTransportEvent`. we have code like
```
const { data } = (event as SnackTransportEvent);
```
so we need the data key here. leaving a comment anyway: https://github.com/expo/snack/pull/376/commits/d515e4dc068c9f33aaab12fc169b272b81a504c9#diff-515958a3d7297a4dd129e86e7a4f99dc5a4d2071958d80bb6479be90ce27c0aaR48

packages/snack-sdk/src/transports/TransportImplPubNub.ts
Kudo: this is the same as the current code: https://github.com/expo/snack/blob/c524f172c960474c5062f8f920898dfe03958364/packages/snack-sdk/src/transports/Transport.ts#L217-L219
we would not enhance pubnub transport in the meantime.

ide:
I'm removing myself as a review from this PR and if you and Cedric is comfortable moving forward with this, let's merge it. I left one comment as a suggestion about naming.

Kudo:
i'm going to merge this pr first as all review comments should be followed up and it's hard to maintain those stocked prs. if there's any other further comments, please let me know and i could do it in a separated pr.

also, this pr only touches snack-sdk, it should be safe to merge.

PR #377, opened 2023-01-03
[snackpub] add snackpub server

Kudo:
# Why

close ENG-7104

## pr series

#376
👉 #377
#378 
#379
#380 
#381

# How

- add socketio server
- add deploy script

# Test Plan

test deployed server


snackpub/.gitignore
```diff
@@ -0,0 +1,10 @@
+*~
+.DS_Store
+
+npm-debug.log
+node_modules/
+
+logs/*.log
+tmp/*
+
+build/
```
byCedric, line 10: Can we consolidate this in the root `.gitignore`? E.g. the `npm-debug.log` and `node_modules` should already be covered by the root `.gitignore`. Just feels a bit cleaner vs adding these files to every workspace (unless it's doing some really special things ofc)

snackpub/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+
+WORKDIR /app/snackpub
+# Build snackpub
+RUN yarn build
+# Reinstall only production dependencies
+RUN yarn install --frozen-lockfile --production
+# Finalize production image
+RUN yarn cache clean
+
+CMD ["node", "."]
```
byCedric, line 63: Maybe it would be nice to create a "production" stage before finalizing this image, to avoid having installed unnecessary system libraries in production. (e.g. like [Snackager](https://github.com/expo/snack/blob/main/snackager/Dockerfile#L47-L69))

```suggestion
# Set up envs
ENV NODE_ENV "production"

ARG DEPLOY_ENVIRONMENT
ENV DEPLOY_ENVIRONMENT ${DEPLOY_ENVIRONMENT}

# Reinstall dependencies to link the monorepo
RUN yarn install --frozen-lockfile --production=false

WORKDIR /app/snackpub
# Build snackpub
RUN yarn build
# Reinstall only production dependencies
RUN yarn install --frozen-lockfile --production
# Finalize production image
RUN yarn cache clean

# Build minimized production image
FROM node:${node_version}-alpine

# Set up envs
ENV NODE_ENV "production"

ARG DEPLOY_ENVIRONMENT
ENV DEPLOY_ENVIRONMENT ${DEPLOY_ENVIRONMENT}

COPY --from=dev /app/snackpub ./snackpub

WORKDIR /app/snackpub

CMD ["node", "."]
```

snackpub/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "snackpub",
+  "version": "0.0.0",
+  "private": true,
+  "description": "The Snack PubSub system",
+  "main": "build/index",
```
byCedric, line 6: Not sure if required, but I don't trust Node to resolve to the exact file 🙈 

```suggestion
  "main": "build/index.js",
```

snackpub/src/RedisAdapter.ts
```diff
@@ -0,0 +1,42 @@
...
+    } catch (e: any) {
+      console.error('Failed to bind redis adapter', e);
+    }
+  }
+}
+
+export async function maybeCloseRedisConnectionsAsync() {
+  subClient?.quit();
+  pubClient?.quit();
+}
```
byCedric, line 42: Do we need to remove the previous client instances too? E.g.

```suggestion
export async function maybeCloseRedisConnectionsAsync() {
  subClient?.quit();
  pubClient?.quit();
  
  subClient = null;
  pubClient = null;
}
```

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  process.once('SIGUSR2', () => shutdown('SIGUSR2'));
+}
+
+function closeServerAsync(server: Server): Promise<void> {
+  return new Promise((resolve) => {
+    server.close((error) => {
+      if (error) {
+        console.log('Failed to close server', error);
+      }
+      resolve();
```
byCedric, line 38: Can this hang for some reason? It's always resolving, even when erroring.

byCedric (approved):
gentleman-club

snackpub/src/index.ts
Kudo: the function is called in the shutdown handler. after this function finishing its work, we can disconnect redis clients. even if the server closing failed, we still have to disconnect redis clients. that's why it's always resolving here.

other than this approach, we can also do this way, but i think this is more complicated.

```ts
function closeServerAsync(server: Server): Promise<void> {
  return new Promise((resolve, reject) => {
    server.close((error) => {
      if (error) {
        reject(error);
      } else {
        resolve();
      }
    });
  });
}

const shutdown = async (signal: ShutdownSignal) => {
  console.log(
    `Received ${signal}; the HTTP server is shutting down and draining existing connections`
  );
  try {
    await closeServerAsync(server);
  } catch (error) {
    console.log('Failed to close server', error);
  }
  await maybeCloseRedisConnectionsAsync();
};
```

.github/workflows/snackpub.yml
```diff
@@ -0,0 +1,184 @@
...
+      - name: 🏗 Setup Google Cloud SDK
+        uses: ./.github/actions/setup-google-cloud
+        with:
+          project-id: exponentjs
+          project-zone: us-central1
+          project-cluster: general-central
+          service-key: ${{ secrets.GCLOUD_KEY }}
+
+      - name: 🛠 Build snackpub
+        run: skaffold build --filename snackpub/skaffold.yaml --file-output snackpub/build.json
```
nicknovitski, line 83: There's no need to use the `--file-output` flag if the deploy script always uses the `latest` tag instead of the contents of `build.json`

nicknovitski (approved):
This looks correct to me.  Newer versions of skaffold support cloud run as a deployment target, but that's a separate change.

Kudo:
> This looks correct to me. Newer versions of skaffold support cloud run as a deployment target, but that's a separate change.

i saw that is still experimental but excited for that. in the current state, using skaffold simply for docker build looks overkilled.
thanks for the neat review 🙏

.github/actions/setup-snackpub/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Setup Snackpub
+description: Prepare Snackpub in GitHub Actions
+
+inputs:
+  node-version:
+    description: Version of Node to use
+    default: 16.x
```
ide, line 7: Is it possible to use Node 18.x since that's the most recent LTS version? (I think Node 16 is fine for this PR, though.)

.github/workflows/snackpub.yml
```diff
@@ -0,0 +1,184 @@
...
+        run: skaffold build --filename snackpub/skaffold.yaml
+        working-directory: ./
+
+  deploy-staging:
+    if: ${{ (github.event.inputs.deploy == 'staging' && github.event_name != 'pull_request') || (github.event_name == 'push' && github.ref == 'refs/heads/main') }}
+    needs: review
+    runs-on: ubuntu-latest
+    environment:
+      name: snackpub-staging
+      url: https://staging.snackpub.expo.io/status
```
ide, line 92: We use expo.dev for new domain names. We also use "staging-X" for our staging domain names because of two reasons:
1. Our wildcard TLS certificate between the load balancer and the CDN is for `*.expo.dev` but not `*.snackpub.expo.dev`.
2. It makes the number of subdomain components the same between staging and production.

```suggestion
      url: https://staging-snackpub.expo.dev/status
```

.github/workflows/snackpub.yml
```diff
@@ -0,0 +1,184 @@
...
+          author_name: Deploy Snackpub to Staging
+          fields: message,commit,author,job,took
+
+  deploy-production:
+    if: ${{ github.event.inputs.deploy == 'production' && github.ref == 'refs/heads/main' }}
+    needs: review
+    runs-on: ubuntu-latest
+    environment:
+      name: snackpub-production
+      url: https://snackpub.expo.io/status
```
ide, line 141:
```suggestion
      url: https://snackpub.expo.dev/status
```

snackpub/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# Prepare builder for production image without any other sub-packages
+FROM node:${node_version}-alpine AS builder
+WORKDIR /app
+
+RUN apk add --no-cache --upgrade \
+    bash \
+    busybox \
+    g++ \
+    make \
+    python3
```
ide, line 34: Do we need all of these dependencies like g++ and python3? (Sometimes we get lucky and don't need them if the npm packages we're installing don't need to natively compile Node modules.)

.github/workflows/snackpub.yml
```diff
@@ -0,0 +1,184 @@
...
+          SKAFFOLD_PUSH_IMAGE: 'true'
+
+      - name: 🚀 Deploy snackpub
+        run: ./scripts/deploy.sh -e production
+        working-directory: snackpub
+
+      - name: 🧹 Prune docker image
+        if: ${{ success() }}
+        run: ./bin/prune-gcr snackpub
+        working-directory: ./
```
ide, line 173: Thank you for adding this

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    await closeServerAsync(server);
+    await maybeCloseRedisConnectionsAsync();
+  };
+
+  // TODO: In Node 9, the signal is passed as the first argument to the listener
+  process.once('SIGHUP', () => shutdown('SIGHUP'));
+  process.once('SIGINT', () => shutdown('SIGINT'));
+  process.once('SIGTERM', () => shutdown('SIGTERM'));
+  // Nodemon sends SIGUSR2 when it restarts the process it's monitoring
+  process.once('SIGUSR2', () => shutdown('SIGUSR2'));
```
ide, line 29:
```suggestion
  process.once('SIGHUP', shutdown);
  process.once('SIGINT', shutdown);
  process.once('SIGTERM', shutdown);
  // Nodemon sends SIGUSR2 when it restarts the process it's monitoring
  process.once('SIGUSR2', shutdown);
```

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+function closeServerAsync(server: Server): Promise<void> {
+  return new Promise((resolve) => {
+    server.close((error) => {
+      if (error) {
+        console.log('Failed to close server', error);
```
ide, line 36: Write to stderr:
```suggestion
        console.error('Failed to close server', error);
```

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  const io = new Server<ClientToServerEvents, ServerToClientEvents, InterServerEvents, SocketData>({
+    serveClient: false,
+  });
+
+  await maybeBindRedisAdapterAsync(io);
+
+  io.on('connection', (socket) => {
+    debug('onconnect', socket.handshake.address);
+    socket.on('message', (data) => {
+      // debug('onMessage', data);
```
ide, line 53:
```suggestion
```

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+      const { channel, message, sender } = data;
+      socket.to(channel).emit('message', { channel, message, sender });
```
ide, line 55: Are channel names unguessable, similar to a session secret? We want to make sure that the only clients who can connect to a channel are the ones that already know about the channel.

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    socket.on('unsubscribeChannel', (data) => {
+      debug('onUnsubscribeChannel', data);
+      const { channel } = data;
+      socket.leave(channel);
+    });
+  });
+
+  io.of('/').adapter.on('join-room', async (channel: string, id: string) => {
+    const sockets = await io.in(channel).fetchSockets();
+    const sender = sockets.filter((socket) => socket.id === id)?.[0]?.data?.deviceId;
```
ide, line 74: The return value of `sockets.filter()` is always an array, and if there is an element in the array, `socket.data` is always defined:

```suggestion
    const sender = sockets.filter((socket) => socket.id === id)[0]?.data.deviceId;
```

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+      if (socket.id !== id) {
+        debug('joinChannel', { channel, sender });
+        socket.emit('joinChannel', { channel, sender });
+      }
+    }
+  });
+
+  io.of('/').adapter.on('leave-room', async (channel: string, id: string) => {
+    const sockets = await io.in(channel).fetchSockets();
+    const sender = sockets.filter((socket) => socket.id === id)?.[0]?.data?.deviceId;
```
ide, line 88:
```suggestion
    const sender = sockets.filter((socket) => socket.id === id)[0]?.data.deviceId;
```

ide (changes requested):

snackpub/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+  });
+}
+
+async function runAsync() {
+  const io = new Server<ClientToServerEvents, ServerToClientEvents, InterServerEvents, SocketData>({
+    serveClient: false,
+  });
+
+  await maybeBindRedisAdapterAsync(io);
```
ide, line 48: Instead of `maybeBindRedisAdapterAsync`, I think it would be clearer to write:
```ts
if (Env.redisURL) {
  await bindRedisAdapterAsync(io);
}
```

This way it's easier to see when the Redis adapter is used, and also to implement logic like:

```ts
if (Env.nodeEnv === 'production') {
  assert(Env.redisURL, `Redis must be configured`);
  await bindRedisAdapterAsync(io);
}
```
^ although, I don't think that's quite the right logic because we may want to test Redis when NODE_ENV is development.

snackpub/src/RedisAdapter.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      pubClient = createClient({ ...options, ...inputOptions });
+      subClient = pubClient.duplicate();
+
+      await Promise.all([pubClient.connect(), subClient.connect()]);
+      server.adapter(createAdapter(pubClient, subClient, { key: 'snackpub' }));
+      console.log('[RedisAdapter] Using redis adapter.');
+    } catch (e: any) {
+      console.error('Failed to bind redis adapter', e);
+    }
```
ide, line 35: IMO we should remove this catch clause and fail to start the server if we can't connect to Redis. We always want to be using Redis in production.

snackpub/src/RedisAdapter.ts
```diff
@@ -0,0 +1,45 @@
...
+    try {
+      const options: Parameters<typeof createClient>[0] = {
+        url: Env.redisURL,
+      };
+      if (Env.redisURL.startsWith('rediss:') && Env.redisTlsCa) {
+        options.socket = {
+          tls: true,
+          ca: fs.readFileSync(Env.redisTlsCa),
+        };
+      }
```
ide, line 25: If the URL starts with rediss: but there is no CA certificate, does the connection to Redis fail? (We do not want to accidentally connect to Redis without encryption, so failing is good.)

ide (commented):
I think we will also need to add a health check so that the load balancer can detect unhealthy instances and also so the load balancer doesn't send more traffic when Redis is unavailable. I'm not sure how health checks work for WebSocket servers - maybe we'll need to add an HTTP endpoint too. For simplicity, I think the health check can be added in a separate PR.

snackpub/src/RedisAdapter.ts
nicknovitski: Won't it not start with that in development?

ide: If we connect to a local Redis instance, I'd expect to use `redis:` (not `rediss:`) and not need a TLS certificate

.github/actions/setup-snackpub/action.yml
Kudo: there're still node 16 in other snack services. let's do node upgrade in a separated pr.

.github/workflows/snackpub.yml
Kudo: i don't have enough permission on expo/snack. let me clarify this after getting enough permission to check this environment usage. maybe we could even drop the environment block.

snackpub/Dockerfile
Kudo: these are necessary for the two packages `bufferutil` and `utf-8-validate` to build native code. these two packages are for websocket performance: https://socket.io/docs/v4/performance-tuning/#install-ws-native-add-ons

snackpub/src/index.ts
Kudo: this is [the current implementation](https://github.com/expo/snack/blob/bd1394bda92c5a45791de13d600707dbf6081bc6/packages/snack-sdk/src/utils.ts#L10-L28) to generate channel. as mentioned on slack, as we have channel rate limiter now. i thought we could enhance the security later.

the channel authentication briefly in my thought is between snack-sdk <-> runtime. it has no difference between snackpub or pubnub.

snackpub/src/index.ts
Kudo: yep the code later is changed as something like that: https://github.com/expo/snack/blob/fd213775d3b2b755e6b15e2e7ff5416d67f3c7fe/snackpub/src/index.ts#L73-L89

i'll add the code accordingly.

```ts
if (Env.nodeEnv === 'production') {
  assert(Env.redisURL, `Redis must be configured`);
}
```

snackpub/src/RedisAdapter.ts
Kudo: > If the URL starts with rediss: but there is no CA certificate, does the connection to Redis fail? (We do not want to accidentally connect to Redis without encryption, so failing is good.)

from my testing for the connection to redislab, it will throw errors.

> Won't it not start with that in development?

by default we don't have redis for snackpub in development, the socketio is running upon in-memory mode. if we have redis running locally with `REDIS_URL` environment variable, then we can test redis mode as well. and yes, `REDIS_URL='redis://localhost:6379'`

snackpub/src/index.ts
ide: Ok, 60 bits should be secure enough because we have the rate limiter and Snack channels do not live long enough for someone to try billions of combinations.

ide (approved):
Approving since my comments about throwing an error if we can't connect to Redis in production are addressed in https://github.com/expo/snack/pull/381.

Let @nicknovitski or @randomhajile know what permissions you need to set up snackpub.expo.dev and staging-snackpub.expo.dev. The configuration for these DNS records will be inside of Universe.

snackpub/skaffold.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: snackpub
+build:
+  platforms: ['linux/amd64']
+  artifacts:
+    - image: gcr.io/exponentjs/snackpub
+      sync:
+        manual:
+          - src: src/**/*
+            dest: /snack/app
+      docker:
```
nicknovitski, line 13: This is all fine; you can also set the required build context in the skaffold yaml itself.

```suggestion
      context: ../
      docker:
```

nicknovitski (approved):

snackpub/skaffold.yaml
Kudo: have to revert the change in https://github.com/expo/snack/pull/377/commits/2728b09d0375a1f3e4d54b54323c2e6976001727. it will transfer context files out from snack root, e.g. `/path/to/snack` in the git repo and the `context: ../` will load the files from `/path/to` which is not quite correct.
will keep the code as-is.

PR #378, opened 2023-01-03
[website] add snackpub a/b testing

Kudo:
# Why

close ENG-7103
close ENG-7107

## pr series

#376
#377
👉 #378 
#379
#380 
#381

# How

- rollout snackpub a/b testing from website. currently 90% pubnub and 10% snackpub + pubnub by default.
- can also add a `testTransport=snackpub` or `testTransport=pubnub` in query string to override the transport from a/b testing randomness.

# Test Plan

- snackpub integration test
- ci passed


byCedric (approved):

website/deploy/production/snack.env
```diff
@@ -1,6 +1,7 @@
 LEGACY_SERVER_URL=https://expo.io
 SERVER_URL=https://expo.dev
 API_SERVER_URL=https://exp.host
+SNACKPUB_URL=https://snackpub-3hn4fth2vq-uc.a.run.app
```
nicknovitski, line 4: These are stable urls?

nicknovitski (approved):

website/deploy/production/snack.env
Kudo: it seems so. between deployments and even delete/recreate the cloud run service, it still sticks on the same name. as the url is not visible to users, if it changed, we could still re-deploy snack with new urls. that's why i didn't put the cloud run service behind google cloud load balancer.

ide: It doesn't need to block this PR but let's set up a load balancer later

ide (commented):
Removing myself as a review; will let others take over for code review of the client-side changes.

PR #379, opened 2023-01-03
[runtime] add snackpub integration

Kudo:
# Why

close ENG-7103
close ENG-7101

## pr series

#376
#377
#378 
👉 #379
#380 
#381

# How

- similiar to snack-sdk, now we have three different transports:
  - RuntimeTransportImplPubNub
  - RuntimeTransportImplWebPlayer
  - RuntimeTrafficMirroringTransport = RuntimeTransportImplPubNub + RuntimeTransportImplSocketIO (for traffic mirror)
- use the initialUrl to determine which transport to use. the website and snack-sdk will append `testTransport` to the initialUrl.
- to prevent duplicated events from the traffic mirror, we implement the deduplication in runtime. the `RuntimeTrafficMirroringTransport` will try `RuntimeTransportImplSocketIO` as the primary transport to send messages or pass messages to upper event handlers. only if the socketio transport is not stable, we will use the RuntimeTransportImplPubNub as fallback.

# Test Plan

- unit tests in runtime
- snackpub integration test
- ci passed


runtime/src/Messaging.tsx
```diff
@@ -1,71 +1,60 @@
...
+  transport = new transportClass(device);
 };
 
-// End existing PubNub subscription, if any
 export const unsubscribe = () => {
-  if (currentChannel) {
-    Logger.comm('Unsubscribing from channel', currentChannel);
-    pubnub.unsubscribe({ channels: [currentChannel] });
+  if (!transport) {
+    throw new Error('Transport does not initialize. Please call `Messaging.init()` first.');
```
byCedric, line 36:
```suggestion
    throw new Error('Transport not initialized, call `Messaging.init()` first.');
```

runtime/src/Messaging.tsx
```diff
@@ -1,71 +1,60 @@
...
-// Initiate PubNub subscription to given `channel`, throw if couldn't connect / timed out. Ends
-// existing subscription, if any.
 export const subscribe = ({ channel }: { channel: string }) => {
-  unsubscribe();
-
-  currentChannel = channel;
-  Logger.comm('Subscribing to channel', channel);
-  pubnub.subscribe({ channels: [channel] });
+  if (!transport) {
+    throw new Error('Transport does not initialize. Please call `Messaging.init()` first.');
```
byCedric, line 43:
```suggestion
    throw new Error('Transport not initialized, call `Messaging.init()` first.');
```

runtime/src/Messaging.tsx
```diff
@@ -1,71 +1,60 @@
...
+  }
+  transport.subscribe(channel);
 };
 
-// Add a message listener
-export const listen = (listener: ListenerParameters['message']) => {
-  pubnub.addListener({ message: listener });
+export const listen = (listener: (payload: RuntimeMessagePayload) => void) => {
+  if (!transport) {
+    throw new Error('Transport does not initialize. Please call `Messaging.init()` first.');
```
byCedric, line 50:
```suggestion
    throw new Error('Transport not initialized, call `Messaging.init()` first.');
```

runtime/src/Messaging.tsx
```diff
@@ -1,71 +1,60 @@
...
-// Publish a message to the currently subscribed channel, if any
 export const publish = (message: object) => {
-  if (currentChannel) {
-    Logger.comm('Sending message', message);
-    pubnub.publish({
-      channel: currentChannel,
-      message: { ...message, device },
-    });
+  if (!transport) {
+    throw new Error('Transport does not initialize. Please call `Messaging.init()` first.');
```
byCedric, line 57:
```suggestion
    throw new Error('Transport not initialized, call `Messaging.init()` first.');
```

byCedric: Also, do we need to throw or can we log this somewhere? Not sure how useful it is if our users are running into this exception?

byCedric (approved):
👍

runtime/src/Messaging.tsx
Kudo: > Also, do we need to throw or can we log this somewhere? Not sure how useful it is if our users are running into this exception?

this is developer error and should not happen. somehow to make typescript happy like `nullthrow(transport)` or `assert(transport)`. since it's running in app runtime, i don't want to introduce more dependencies for nullthow or assert.

byCedric: That makes sense 👍

runtime/src/UrlUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+  const channel = matches[2];
+
+  let testTransport = null;
+  const queryItems = (matches[3] ?? '').split(/&/g);
+  for (const item of queryItems) {
+    if (item.startsWith('testTransport=')) {
+      testTransport = item.substring(14);
+      break;
+    }
+  }
```
ide, line 17: Using the built-in URL API for parsing might be easier. It has been supported for about 7-8 years: https://caniuse.com/?search=URL

ide (commented):
Similar to the other client PRs, I am removing myself as a reviewer and leaving it to you and Cedric to decide if you are comfortable moving forward with merging the code and being responsible for it.

PR #380, opened 2023-01-05
[snack-sdk][website] add snackpub metrics

Kudo:
# Why

close ENG-7102

## pr series

#376
#377
#378 
#379
👉 #380 
#381

# How

- [website] Add two snackpub metrics for connection quality analytics
  - `TRANSPORT_CONNECTION_SUCCEEDED` with elapsed time ms and attempt times
  - `TRANSPORT_CONNECTION_FAILED` with elapsed time ms and attempt times
- we don't have analytics for runtime, so runtime is not covered by this metrics

# Test Plan

snackpub integration test

PR #381, opened 2023-01-05
[snack-sdk][website][runtime] add rate limiter

Kudo:
# Why

close ENG-7105

## pr series

#376
#377
#378 
#379
#380 
👉 #381

# How

add rate limiter:
  - 30 connections per IP per hour
  - 600 messages per socket.id per second and 6000 messages per socket.id per minute
  - 30 channels per socket.id per minute

# Test Plan

snackpub manual test

### DIFF
--- a/snackpub/src/RateLimiter.ts
+++ b/snackpub/src/RateLimiter.ts
@@ -11,7 +11,7 @@ export default class RateLimiter {
   public hasExceededSrcIpRateAsync(
     srcIp: string,
     _socketId: string,
-    options: RateLimiterOptions = { maxOperations: 30, intervalSeconds: 60 } // 30 connections per IP per hour
+    options: RateLimiterOptions = { maxOperations: 30, intervalSeconds: 60 } // 30 connections per IP per minute
   ): Promise<boolean> {
     return this.hasExceededKeyAsync(srcIp, `ip:${srcIp}`, options);
   }
@@ -45,7 +45,7 @@ export default class RateLimiter {
       .expire(redisKey, intervalSeconds)
       .exec();
 
-    const operationsInt = parseInt((operations ?? '0').toString(), 10);
+    const operationsInt = parseInt(operations ?? '0', 10);
     // The `operationsInt` is the value after incr(), we should minus one back
     if (operationsInt - 1 >= options.maxOperations) {
       await this.redisClient.decr(redisKey);
@@ -56,7 +56,7 @@ export default class RateLimiter {
   }
 
   private createRedisKey(key: string, intervalSeconds: number): string {
-    const intervalIndex = Math.floor(new Date().getTime() / 1000 / intervalSeconds);
+    const intervalIndex = Math.floor(Date.now() / 1000 / intervalSeconds);
     // Simplified rate limiter that cache hitting doesn't extend the time window.
     return `snackpub:rate:${key}:${intervalIndex}`;
   }
--- a/snackpub/src/RateLimiter.ts
+++ b/snackpub/src/RateLimiter.ts
@@ -16,12 +16,26 @@ export default class RateLimiter {
     return this.hasExceededKeyAsync(srcIp, `ip:${srcIp}`, options);
   }
 
-  public hasExceededMessagesRateAsync(
+  public async hasExceededMessagesRateAsync(
     srcIp: string,
     socketId: string,
-    options: RateLimiterOptions = { maxOperations: 600, intervalSeconds: 1 } // 600 messages per socket.id per second
+    options?: {
+      dosPrevention?: RateLimiterOptions;
+      fairUsage?: RateLimiterOptions;
+    }
   ): Promise<boolean> {
-    return this.hasExceededKeyAsync(srcIp, `messages:${socketId}`, options);
+    // A short internal to prevent DOS: 600 messages per socket.id per second
+    const optsDosPrevention = options?.dosPrevention ?? { maxOperations: 600, intervalSeconds: 1 };
+
+    // Longer internal for fair usage: 6000 messages per socket.id per minute
+    const optsFairUsage = options?.fairUsage ?? { maxOperations: 6000, intervalSeconds: 60 };
+
+    const results = await Promise.all([
+      this.hasExceededKeyAsync(srcIp, `messagesDosPrevention:${socketId}`, optsDosPrevention),
+      this.hasExceededKeyAsync(srcIp, `messagesFairUsage:${socketId}`, optsFairUsage),
+    ]);
+
+    return results.some((result) => !!result);
   }
 
   public hasExceededChannelsRateAsync(
@@ -45,11 +59,13 @@ export default class RateLimiter {
       .expire(redisKey, intervalSeconds)
       .exec();
 
-    const operationsInt = parseInt(operations ?? '0', 10);
-    // The `operationsInt` is the value after incr(), we should minus one back
-    if (operationsInt - 1 >= options.maxOperations) {
-      await this.redisClient.decr(redisKey);
+    const operationsString = operations?.toString() ?? '0';
+    const operationsInt = parseInt(operationsString, 10);
+    if (operationsInt === options.maxOperations + 1) {
+      // Log only once when exceeding limits
       console.log(`[RateLimiter] exceeding limits - srcIp[${srcIp}] redisKey[${redisKey}]`);
+    }
+    if (operationsInt > options.maxOperations) {
       return true;
     }
     return false;
--- a/snackpub/src/RateLimiter.ts
+++ b/snackpub/src/RateLimiter.ts
@@ -1,0 +1,63 @@
+import type { createClient } from 'redis';
+
+interface RateLimiterOptions {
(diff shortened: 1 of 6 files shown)


snackpub/src/index.ts
```diff
@@ -40,23 +47,65 @@ function closeServerAsync(server: Server): Promise<void> {
...
+function terminateSocket(
+  socket: Socket<ClientToServerEvents, ServerToClientEvents, InterServerEvents, SocketData>,
+  reason: string
+) {
+  socket.emit('terminate', reason);
+  setTimeout(() => {
+    if (socket.connected) {
+      socket.disconnect(true);
+    }
+  }, 3000);
```
byCedric, line 59: Does this 3s timeout has a specific meaning here? If so, might be nice to add a comment on why 3s was chosen

Kudo: added a comment

```
  // When clients receive the `terminate` message, they should disconnect and no longer try to reconnect back to the server.
  // In case the clients do not handle the `terminate` message, e.g. from other websocket clients other than snackpub clients.
  // We will wait three seconds to close the connections actively.
```

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+  intervalSeconds: number;
+}
+
+export default class RateLimiter {
+  constructor(private readonly redisClient: ReturnType<typeof createClient>) {}
+
+  public hasExceededSrcIpRateAsync(
+    srcIp: string,
+    _socketId: string,
+    options: RateLimiterOptions = { maxOperations: 30, intervalSeconds: 60 } // 30 connections per IP per hour
```
ide, line 14:
```suggestion
    options: RateLimiterOptions = { maxOperations: 30, intervalSeconds: 60 } // 30 connections per IP per minute
```

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+  ): Promise<boolean> {
+    return this.hasExceededKeyAsync(srcIp, `ip:${srcIp}`, options);
+  }
+
+  public hasExceededMessagesRateAsync(
+    srcIp: string,
+    socketId: string,
+    options: RateLimiterOptions = { maxOperations: 600, intervalSeconds: 1 } // 600 messages per socket.id per second
```
ide, line 22: 600 messages per second per source IP seems like a lot. What is the use case for this many messages?

We could also set up two rate limits: 600/second and 6000/minute. This would allow for high numbers of messages at once, but not continuously.

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+  ): Promise<boolean> {
+    const { intervalSeconds } = options;
+    const redisKey = this.createRedisKey(key, intervalSeconds);
+    const [operations] = await this.redisClient
+      .multi()
+      .incr(redisKey)
+      .expire(redisKey, intervalSeconds)
+      .exec();
+
+    const operationsInt = parseInt((operations ?? '0').toString(), 10);
```
ide, line 48: Shouldn't the value always be a string or null?
```suggestion
    const operationsInt = parseInt(operations ?? '0', 10);
```

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+    // The `operationsInt` is the value after incr(), we should minus one back
+    if (operationsInt - 1 >= options.maxOperations) {
```
ide, line 50:
```suggestion
    if (operationsInt > options.maxOperations) {
```

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+      await this.redisClient.decr(redisKey);
```
ide, line 51: I don't think we need to do this. Firstly, if we can avoid extra traffic to Redis that's slightly better for performance. Also, I don't think it matters if the counter keeps growing infinitely.

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+      console.log(`[RateLimiter] exceeding limits - srcIp[${srcIp}] redisKey[${redisKey}]`);
```
ide, line 52: What do you think about making this a debug log message? I am a little worried about our logs becoming filled with messages. Or, we could write logic like `if (operationsInt === options.maxOperations + 1) { console.log() }` to log just one time.

snackpub/src/RateLimiter.ts
```diff
@@ -0,0 +1,63 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  private createRedisKey(key: string, intervalSeconds: number): string {
+    const intervalIndex = Math.floor(new Date().getTime() / 1000 / intervalSeconds);
```
ide, line 59: Don't need to allocate a new Date object:
```suggestion
    const intervalIndex = Math.floor(Date.now() / 1000 / intervalSeconds);
```

ide (commented):
I didn't review all the code yet but wanted to leave some comments early.

snackpub/src/index.ts
```diff
@@ -40,23 +47,69 @@ function closeServerAsync(server: Server): Promise<void> {
...
+  socket.emit('terminate', reason);
+
+  // When clients receive the `terminate` message, they should disconnect and no longer try to reconnect back to the server.
+  // In case the clients do not handle the `terminate` message, e.g. from other websocket clients other than snackpub clients.
+  // We will wait three seconds to close the connections actively.
+  setTimeout(() => {
+    if (socket.connected) {
+      socket.disconnect(true);
+    }
+  }, 3000);
```
ide, line 63: While it's not necessary, is there a way to listen to disconnection events from the socket and call `clearTimeout` if the client disconnects? That way we don't have unnecessary setTimeout handles.

snackpub/src/RedisAdapter.ts
```diff
@@ -1,42 +1,19 @@
...
-      subClient = pubClient.duplicate();
-
-      await Promise.all([pubClient.connect(), subClient.connect()]);
-      server.adapter(createAdapter(pubClient, subClient, { key: 'snackpub' }));
-      console.log('[RedisAdapter] Using redis adapter.');
-    } catch (e: any) {
-      console.error('Failed to bind redis adapter', e);
-    }
-  }
+  subClient = redisClient.duplicate();
```
ide, line 11: Should this function throw an assertion error if subClient is already defined? That seems like a good thing to do since we don't want to have an orphaned Redis client that hasn't been shut down with `quit()`.

snackpub/src/index.ts
```diff
@@ -12,13 +15,17 @@ import type {
...
+  redisClient: ReturnType<typeof createClient> | null
+) {
   const shutdown = async (signal: ShutdownSignal) => {
     console.log(
       `Received ${signal}; the HTTP server is shutting down and draining existing connections`
     );
     await closeServerAsync(server);
-    await maybeCloseRedisConnectionsAsync();
+    await closeRedisAdapterAsync();
+    await redisClient?.quit();
```
ide, line 28: I think this shutdown routine needs to be more graceful. Specifically, we need to reject all new connections so that we drain connections from the server. This is where a readiness probe or health check is especially helpful. A failing health check tells the load balancer to stop sending new traffic to an unhealthy server instance. Additionally, we should notify the clients that we're going to terminate the connections and that they should try to reconnect to another instance. This can happen in another PR but I think it is important for service reliability.

That said, I think the code here does a good job of making sure to stop the WebSocket server before terminating the connections to Redis. We could also use Promise.all to terminate both Redis connections simultaneously.

snackpub/src/index.ts
```diff
@@ -12,13 +15,17 @@ import type {
...
   };
 
   // TODO: In Node 9, the signal is passed as the first argument to the listener
```
ide, line 31: Unrelated: can remove this TODO comment

snackpub/src/index.ts
```diff
@@ -40,23 +47,69 @@ function closeServerAsync(server: Server): Promise<void> {
...
+    redisClient = createClient(redisClientOptions);
+    await redisClient.connect();
+
+    await bindRedisAdapterAsync(io, redisClient);
+    rateLimiter = new RateLimiter(redisClient);
+  }
 
-  io.on('connection', (socket) => {
+  io.on('connection', async (socket) => {
     debug('onconnect', socket.handshake.address);
```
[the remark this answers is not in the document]
Kudo: will follow up in a separated pr and test whether we should get the ip from the `X-Forwarded-For` or somewhat else that GCLB provided.

snackpub/src/index.ts
Kudo: will follow up readiness in a separated pr.

> We could also use Promise.all to terminate both Redis connections simultaneously.
kinda afraid to do it simultaneously, so to terminate the subscriber client first and then the main redis client.

Kudo:
i'm going to merge this pr and leave some action items in separated prs or tasks.

PR #384, opened 2023-01-20
[website] Mount health endpoint earlier

nicknovitski:
# Why

Currently rollouts are failing, because requests to `/ready` are being answered by a `*` route that is `use`d first.

This change fixes that.

# Test Plan

This can be tested by running `skaffold dev --kube-context=minikube --filename website/skaffold.yaml`.


ide (changes requested):
Instead of a separate port, how about we register the middleware earlier, like after the compress() or sw() middleware? That way we also test that the actual, real Koa instance (most importantly, the HTTP server object) is listening and ready.

nicknovitski:
That's a good point, I hesitated to do that since it would technically change the behavior of the server and I thought this was less invasive, but 👍 .

website/deploy/development/deployment-probes.yaml
```diff
@@ -1,15 +0,0 @@
...
-metadata:
-  name: snack
-spec:
-  template:
-    spec:
-      containers:
-        - name: snack
-          readinessProbe:
-            initialDelaySeconds: 15
-            timeoutSeconds: 20
```
ide, line 12: Shouldn't we keep some probes so that k8s knows to restart a container when it's unhealthy? Separately, I've been thinking that for services that don't have liveness probes, we should still have startup probes so that rollouts fail when there's a fatal error upon startup.

nicknovitski: This file is a patch which modifies the probe stanzas; removing it only means that those stanzas are identical in the development kustomization, which is what we want whenever possible.  At times, a development server might take so long to start that probes must be different in a development environment, but I didn't find that to be the case (maybe it used to be). You can check by running `skaffold dev` on the branch.

For the separate question, any fatal error will prevent the new pods from becoming ready, which will cause the rollout to hang, and then fail after `progressDeadlineSeconds`.  [The only use of a startup probe is to enable having fast-running liveness probes on a slow-starting container](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes).

ide: Looks good to me. I see base/deployment.yaml now.

ide (approved):

PR #386, opened 2023-02-08
[snackager] Switch redis over to memorystore

wschurman:
# Why

To migrate snackager over to memorystore we need a proxy into the VPC containing the memorystore instances for local development. This is the same as the other services (www, push-hydrant).

The difference is that snackager lives in a repo outside of universe, and that repo doesn't have any of the dependencies necessary for running this proxy script or the proxy itself (gcloud). So we had a few options:
1. Move snackager back to universe
2. Add the dependencies to the snack repo
3. Add a script in the universe repo that must be run in tandem with local snackager in order to develop snackager locally

I made an attempt at option 3 but the discussion on the PR settled on option 2 being the better option.

That being said, we use `./install-tools` to set up gcloud in universe and we don't have an equivalent here, so I just added it to the CONTRIBUTING.md guide.

# How

Copied the proxy script from universe into this repo and make it run as part of `yarn start`.

# Test Plan

Run `yarn start` and ensure it works.


snackager/k8s/base/deployment.yaml
```diff
@@ -21,9 +21,6 @@ spec:
         secret:
           secretName: git-account-credentials
           defaultMode: 0600
-      - name: redislabs-tls
-        secret:
-          secretName: redislabs-tls
```
ide, line 26: After deploying this, I think we'll need to manually delete the k8s Secret resource. It's not a big deal if we don't since it's not actually a sensitive value (public certificate) but good to clean up.

snackager/scripts/dev.sh
```diff
@@ -0,0 +1,23 @@
...
+
+port_forward_redis() {
+  # Port-forward Redis in a subshell and prefix its output to be more discernible
+  $DIR/port-forward-redis --environment "$REDIS_ENVIRONMENT" \
+    > >(sed "s/^/${yellow}[Redis]${reset} /") \
+    2> >(sed "s/^/${yellow}[Redis]${reset} /" >&2)
+}
+
+echo
+echo "${yellow}Starting the redis port tunnel and waiting for it to become available${reset}"
```
ide, line 15:
```suggestion
echo "${yellow}Starting the Redis port tunnel and waiting for it to become available${reset}"
```

snackager/scripts/dev.sh
```diff
@@ -0,0 +1,23 @@
...
+
+port_forward_redis
+
+echo
+echo "${yellow}Starting the dev server${reset}"
```
ide, line 20:
```suggestion
echo "${yellow}Starting the Snackager server${reset}"
```

snackager/scripts/port-forward-redis
```diff
@@ -0,0 +1,52 @@
...
+case "$environment" in
+  staging)
+    snackager_cache_memorystore_ip="10.81.199.227"
+    gcloud compute ssh --zone us-central1-f memorystore-redis-proxy -- -N -L 6379:$snackager_cache_memorystore_ip:6379 &
+    until nc -w 10 -z localhost 6379; do sleep 1; done
+    ;;
+  production)
+    snackager_cache_memorystore_ip="10.144.221.51"
+    gcloud compute ssh --zone us-central1-f memorystore-redis-proxy -- -N -L 6379:$snackager_cache_memorystore_ip:6379 &
+    until nc -w 10 -z localhost 6379; do sleep 1; done
```
ide, line 46: Since these Bash commands are somewhat complex, I wonder if it'd be better to structure the code like:
```
if staging:
  set snackager_cache_memorystore_ip to staging ip
else if production:
  set snackager_cache_memorystore_ip to prod ip
else
  exit with error

do the connection logic with snackager_cache_memorystore_ip now defined
```

(but I could also see the Bash code being messier this way too, not sure it's necessarily better)

ide (changes requested):

snackager/scripts/port-forward-redis
wschurman: kept the switch but factored out the code

ide (approved):

ide:
Thank you for seeing this whole migration through!

PR #390, opened 2023-02-14
chore: use expo sdk 47 by default

byCedric:
# Why

We are preparing to add SDK 48, this wasn't done yet.

# How

Updated the packages and config

# Test Plan


byCedric:
Not sure why CI isn't kicking in for the newer commit, but here is a manual run for the packages: https://github.com/expo/snack/actions/runs/4176328643

PR #396, opened 2023-03-11
[snackpub] redis integration enhancement

Kudo:
# Why

enhance redis integration for snackpub

# How

- make sure to use redis adapter in production env
- handle redis connection errors and reconnection

# Test Plan

ci passed

### DIFF
--- a/snackpub/src/index.ts
+++ b/snackpub/src/index.ts
@@ -169,7 +169,9 @@ function registerRedisClientErrorHandlers(redisClients: NullableRedisClientType[
       debug('Redis client error', error);
     });
     redisClient?.on('reconnecting', () => {
-      debug(`Reconnecting Redis server in ${REDIS_RECONNECT_RETRY_DELAY_MS}ms`);
+      debug(
+        `Reconnecting to Redis server in ${REDIS_RECONNECT_RETRY_DELAY_MS}ms - redisURL[${Env.redisURL}]`
+      );
     });
   }
 }
--- a/snackpub/src/index.ts
+++ b/snackpub/src/index.ts
@@ -163,7 +163,7 @@ function registerShutdownHandlers(server: Server, redisClients: NullableRedisCli
  */
 function registerRedisClientErrorHandlers(redisClients: NullableRedisClientType[]) {
   for (const redisClient of redisClients) {
-    redisClient?.on('error', async (error) => {
+    redisClient?.on('error', (error) => {
       // node-redis emits errors even when reconnecting,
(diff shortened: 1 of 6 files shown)


snackpub/src/index.ts
```diff
@@ -160,3 +136,89 @@ async function runAsync() {
...
+    });
+  }
+}
+
+/**
+ * Gracefully shutdown the Snackpub server
+ */
+async function shutdownAsync(server: Server, redisClients: NullableRedisClientType[]) {
+  await closeServerAsync(server);
+  await Promise.all(redisClients.map((redisClient) => redisClient?.quit));
```
ide, line 182: Need to make sure `quit()` is called:
```suggestion
  await Promise.all(redisClients.map((redisClient) => redisClient?.quit()));
```

snackpub/src/index.ts
```diff
@@ -160,3 +136,89 @@ async function runAsync() {
...
+    });
+  }
+}
+
+/**
+ * Register Redis client handlers
+ */
+function registerRedisClientErrorHandlers(redisClients: NullableRedisClientType[]) {
+  for (const redisClient of redisClients) {
+    redisClient?.on('error', async (error) => {
```
ide, line 166: Looks like this doesn't need to be async?
```suggestion
    redisClient?.on('error', (error) => {
```

snackpub/src/index.ts
```diff
@@ -160,3 +136,89 @@ async function runAsync() {
...
+      // node-redis emits errors even when reconnecting,
+      // leave this error handler as no-op and handle in `reconnectStrategy`.
+      debug('Redis client error', error);
+    });
+    redisClient?.on('reconnecting', () => {
+      debug(`Reconnecting Redis server in ${REDIS_RECONNECT_RETRY_DELAY_MS}ms`);
```
ide, line 172:
```suggestion
      debug(`Reconnecting to Redis server in ${REDIS_RECONNECT_RETRY_DELAY_MS}ms`);
```

Also, it might be helpful to include information about which Redis server we're reconnecting to, if that info is easily available.

ide (changes requested):
Looks mostly good to me

ide (approved):
Like your other PR, I'm approving this one too in order to help keep you unblocked but please address the comments first. I think the comments are mostly straightforward and do not require another round of code review.

Kudo:
good catch for those errors. thanks James as always! all comments in this pr are followed up as well and going to merge this one.

PR #397, opened 2023-03-11
[snackpub] read correct remote address behind load balancer

Kudo:
# Why

follow up https://github.com/expo/snack/pull/381#discussion_r1080836193

# How

- read the address from the `X-Forwarded-For` first then the socket remote address
- because cloud run doesn't log the remote address as http access log, this pr also adds remote address to log.

# Test Plan

- add `getRemoteAddress` unit test
- ci passed

### DIFF
--- a/.github/workflows/snackpub.yml
+++ b/.github/workflows/snackpub.yml
@@ -59,6 +59,9 @@ jobs:
       - name: 🚨 Lint snackpub
         run: yarn lint --max-warnings 0
 
+      - name: 🧪 Unit test website
+        run: yarn test --ci --maxWorkers 1
+
   build:
     if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
--- a/.github/workflows/snackpub.yml
+++ b/.github/workflows/snackpub.yml
@@ -60,7 +60,7 @@ jobs:
         run: yarn lint --max-warnings 0
 
       - name: 🧪 Unit test website
-        run: yarn test --ci --maxWorkers 1
+        run: yarn test --ci
 
   build:
     if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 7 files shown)


snackpub/src/NetworkUtils.ts
```diff
@@ -0,0 +1,12 @@
+import type { IncomingHttpHeaders, IncomingMessage } from 'http';
+
+type ValueOf<T> = T[keyof T];
+
+export function getRemoteAddress(req: IncomingMessage): string | undefined {
+  return getFirstHeaderValue(req.headers['x-forwarded-for']) ?? req.socket.remoteAddress;
```
ide, line 6: Look at server/www/src/middleware/ip.ts. I think it would be a good idea to proactively add support for cf-connecting-ip if we were to add Cloudflare proxies. (I'm just worried that we might add Cloudflare proxies and then break the rate limiter because all of the requests will appear to come from a small set of the same IP addresses owned by the proxy servers.)

.github/workflows/snackpub.yml
```diff
@@ -59,6 +59,9 @@ jobs:
       - name: 🚨 Lint snackpub
         run: yarn lint --max-warnings 0
 
+      - name: 🧪 Unit test website
+        run: yarn test --ci --maxWorkers 1
```
ide, line 63: Any reason --maxWorkers can't be higher? (Or just omit the option?)

ide (changes requested):

ide (approved):
Approving this PR to keep you unblocked but would appreciate it if you could address the comments by either updating the code and including test cases, or explaining why you decided not to update the code.

.github/workflows/snackpub.yml
Kudo: i just copied the --maxWorkers from other snack github jobs and don't know why actually. i think we don't have to keep this for snackpub github jobs. let's removed this.

Kudo:
> Approving this PR to keep you unblocked but would appreciate it if you could address the comments by either updating the code and including test cases, or explaining why you decided not to update the code.

all comments are followed up. thanks for the neat review. i'm going to merge this pr.

PR #408, opened 2023-04-01
[website] Update Readme.md file defaults

amandeepmittal:
# Why

Currently:
- The `Readme.md` file doesn't mention Discord community
- The docs link (even though redirect works) is outdated

CleanShot 2023-04-01 at 22 48 21@2x

# How

This PR:
- Adds a link to the Discord community
- Updates the docs link
- Fixes how we use "Expo CLI" as a term (we're now using `expo-cli` for deprecated global CLI)

# Test Plan

N/A

PR #418, opened 2023-04-18
chore(website): disable random failing website config check

byCedric:
# Why

See the failure of #408

# How

Disabled the failing job by commenting out.

# Test Plan

Rebase PR #408, check if it passes.


byCedric:
cc @nicknovitski, I think you added it. I have no idea what this would mean, or how to fix it. We need to merge Aman's PR #408 so I'm (temporarily) disabling the check.

PR #423, opened 2023-05-08
[sdk] fix code diff limit for snackpub transport

Kudo:
# Why

We have 32768 limit for code diff for PubNub transport and the `32768` limit is hardcoded in different places. I didn't implement similar logic in socketio transport, so it will actually transfer `'X'.repeat(32768)` as the file content in transport. 

# How

- implement 32768 limit on TransportImplSocketIO
- centralize the 32768 limit

# Test Plan

- working example: https://staging-snack.expo.dev/@kudochien/long-snack (since i've deployed the pr to staging)
- not working example: https://snack.expo.dev/@kudochien/long-snack

### DIFF
--- a/packages/snack-sdk/CHANGELOG.md
+++ b/packages/snack-sdk/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unpublished
 
+### 🐛 Bug fixes
+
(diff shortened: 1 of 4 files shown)


byCedric (approved):
LGTM!

PR #431, opened 2023-06-08
fix: vendor `@react-navigation` packages for `expo-router`

byCedric:
# Why

Vendoring `@react-navigation` should allow us to compensate for issues with Expo Router's overrides of the package. With this, we have a single reference of `@react-navigation` packages for all parts of the system.

# How

- Added externals to Snackager
- Installed dependencies and added links in Runtime
- Updated `snack-content` to include the bundled modules

# Test Plan

Pushing to staging to see if it works.


byCedric:
Superseded by #436

PR #436, opened 2023-06-21
feature(runtime): vendor Expo Router to enable partial usage

byCedric:
# Why

This adds `expo-router` to the Snack Runtime, with some workarounds embedded. It allows users to (partially) use Expo Router in Snack environments.

Unfortunately, we are limited in some areas (using Expo Go, other Snack limitations), full usage of Expo Router is not possible (yet). E.g. `<Head>` doesn't work, handoff, envvar customization, etc

# How

- Vendored `expo-router` in Runtime
- Added simplified `expo-router/entry` code that's compatible with Snack Runtime

# Test Plan

Will deploy to staging.


runtime/src/aliases/common.tsx
```diff
@@ -38,6 +38,15 @@ const aliases: { [key: string]: any } = {
...
   '@shopify/react-native-skia/lib/module/web': SkiaWeb,
+
+  // Only works when vendored into the runtime (expo-router@1.5.3)
+  'expo-router': require('expo-router'),
+  'expo-router/stack': require('expo-router/stack'),
+  'expo-router/tabs': require('expo-router/tabs'),
+  'expo-router/drawer': require('expo-router/drawer'),
+  'expo-router/html': require('expo-router/html'),
+  'expo-router/head': require('expo-router/head'),
+  'expo-router/entry': () => {}, // noop
```
byCedric, line 49: Provide a warning to the user when using this

runtime/src/App.tsx
```diff
@@ -437,19 +438,26 @@ export default class App extends React.Component<object, State> {
...
 
-      const hasRootModuleUri = await Modules.has(rootModuleUri);
-      if (!hasRootModuleUri) {
-        const rootDefaultExport = (await Modules.load(rootModuleUri)).default;
-        if (!rootDefaultExport) {
-          throw new Error(`No default export of '${Files.entry()}' to render!`);
+      // Special handling for Expo Router projects
+      if (Modules.hasDependency('expo-router')) {
+        // NOTE(cedric): this is the default `require.context('./app')` virtual module equivalent, see `snack-require-context`
+        const ctx = await Modules.load('module://app?ctx=eyJyIjp0cnVlLCJtIjoiLioiLCJvIjoic3luYyJ9');
```
Kudo, line 448: nix: maybe we could move the `module://app?ctx=eyJyIjp0cnVlLCJtIjoiLioiLCJvIjoic3luYyJ9` as an export constant from `ExpoRouterEntry`. it would be clear if we put all things together.

Kudo (approved):
just a nit comment and everything looks good to me.

runtime/src/App.tsx
byCedric: Sounds good

PR #441, opened 2023-07-12
fix(runtime): improve virtual module invalidation

byCedric:
# Why

This fixes a couple of things around the newer `require.context` implementation for Expo Router.

# How

- Keep track of each dependent of a virtual module, invalidate only those virtual modules when dependent was changed.
- Invalidate all virtual modules whenever a new file was added
- Only flush the root component when not using `expo-router`

# Test Plan

See staging.


byCedric:
This seems to work way better than what we had.

PR #442, opened 2023-07-17
[infra] Deploy snackpub to Kubernetes.

randomhajile:
# Why

We've decided that deploying snackpub to our own Kubernetes infrastructure is more efficient than deploying to Cloud Run. This PR will allow us to deploy to both environment so that we can start pointing snackpub traffic from Cloud Run to k8s.

# How

- Added k8s config for both staging and production snackpub.
- Updated Skaffold config to allow deploying to k8s.
- Updated snackpub GHA workflow to deploy to k8s after deploying to Cloud Run.

# Test Plan

- I ran `skaffold build` and `skaffold deploy` on my machine to check that both work.
- Once merged, I'll watch the deploy actions to make sure they work.


nicknovitski (approved):
Are the secret values duplicated to a new file here, or are they reused somehow?

byCedric (approved):
> I ran `skaffold build` and `skaffold deploy` on my machine to check that both work.

Looks good to me, and if it doesn't work, we'll just ship your machine instead.

PR #447, opened 2023-07-25
[website] make emotions more neutral

kbrandwijk:
# Why

I noticed a few negative outliers in the range of possible emotions a Snack can have.

# How

Removed a few of the ones that can invoke negative associations.

# Test Plan

Didn't test. It's a list of strings...


kbrandwijk:
87x75 (6525) down to 73x75 (5475) isn't huge, but definitely not negligible. That being said, I think the number of possibilities is still large enough though, they don't need to be globally unique, right?

byCedric:
Unfortunately, they do need to be globally unique. So let's find something to bump it up to at least 6000+

We could add something boring like "weather" (`sunny`, `rainy`, `windy`) etc as additional set, that should boost it to the safe zone.

byCedric:
Color seems like a good option too yeah

byCedric:
Merging this in, but will add color.

PR #452, opened 2023-08-07
feature(website): add support for opening projects with Orbit

gabrieldonadel:
# Why

As a part of the new Expo Orbit project, we need to add a button to the `EditorToolbar` in order to
 allow users to launch their projects with just one click using Expo Orbit.

# How

- Add "Open with Orbit" button to the `EditorToolbar` using the `expo-orbit://` custom schema
- Adds `custom-protocol-check` dependency to determine if Expo Orbit is installed or not and prompt user to install it if necessary
- Update AuthManager to query `experiments` to determine if we should display the Orbit button 

# Test Plan

### Case 1

The user HAS Obit installed and presses the "Open with Orbit" button  
 
https://github.com/expo/snack/assets/11707729/8146d57d-2f5a-4262-a7a4-31cd1b7f21c5

### Case 2
 
The user DOES NOT HAVE Obit installed, and presses the "Open with Orbit" button 

https://github.com/expo/snack/assets/11707729/2ea85073-56fb-4bb9-b102-c4a43552674f

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6293,6 +6293,11 @@ currently-unhandled@^0.4.1:
   dependencies:
     array-find-index "^1.0.1"
 
+custom-protocol-check@^1.4.0:
+  version "1.4.0"
+  resolved "https://registry.yarnpkg.com/custom-protocol-check/-/custom-protocol-check-1.4.0.tgz#87b579f6519d48adf55f539d774f7e026bae35e7"
+  integrity sha512-eMTyp8AKnE5eo+mKNqG3743eb5ZND5LhBgf9F8BN2tVdhSBnOCHH7me7iTcv0BUDhUW2dBQiHWLWMy776yZW1A==
+
 cycle@1.0.x:
   version "1.0.3"
   resolved "https://registry.yarnpkg.com/cycle/-/cycle-1.0.3.tgz#21e80b2be8580f98b468f379430662b046c34ad2"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -14173,11 +14178,35 @@ snack-babel-standalone@^2.2.2:
   resolved "https://registry.yarnpkg.com/snack-babel-standalone/-/snack-babel-standalone-2.2.2.tgz#2154855c482c112036261171ed32f7892102be60"
   integrity sha512-kYYfIsktDfJeYrByF184YO/x55U7rPzfXThsownEGCNBn8d1xZ8AD7NyuLS5aA1n2rk9yKQ1CZxIRHwoNkootQ==
 
+snack-content@~1.3.0:
```
byCedric, line 14181: Hmm, these shouldn't change in the lockfile

gabrieldonadel: I think these are unrelated TBH, I tested deleting my `node_modules` and running `yarn install` from `main` and `yarn.lock` would still change

byCedric: Oh wait, what happens when you run these commands in order from the root of the project?

- `yarn install`
- `yarn build`
- `yarn install`

byCedric: Let's merge it, I can check before deploying.

gabrieldonadel: > Oh wait, what happens when you run these commands in order from the root of the project?
> 
> * `yarn install`
> * `yarn build`
> * `yarn install`

Same thing, unfortunately

PR #453, opened 2023-08-08
feature(website): Update AuthenticationManager to get user data from the GraphQL API

gabrieldonadel:
# Why

In order to support fetching more complex data involving user fields, we should use the GraphQL API instead of `api/v2`. Necessary for querying things like:

```
{
  me {
    id
    username 
    userExperiments {
      id
      enabled
      experiment
    }
  }
}
```

In a follow-up PR we should add something like `graphql-tag` to have more robust support for the GraphQL queries 

# How

Replace the existing `userInfo` request with an equivalent query using the GraphQL API

# Test Plan

Run the website locally and ensure user data is loaded correctly 
image

### DIFF
--- a/website/src/client/auth/authManager.tsx
+++ b/website/src/client/auth/authManager.tsx
@@ -3,24 +3,10 @@ import ExtendableError from 'es6-error';
 import { getAuthStorageKey } from './config';
 import Storage from './storage';
 
-type Auth0UserData = {
-  name: string;
-  username: string;
+export type UserData = {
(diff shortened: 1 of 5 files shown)


byCedric (approved):
Seems good to me. Did you check if the other properties aren't used anywhere?

gabrieldonadel:
> Did you check if the other properties aren't used anywhere?

Yes, the only properties used were id, username and picture

PR #457, opened 2023-08-25
fix(runtime): use `transformer.publicPath` with deploy workaround for asset paths on web

byCedric:
# Why

When importing assets from packages bundled with the Snack Runtime, the assets are expected to be hosted at the root of the domain. This adds the `transformer.publicPath` to tell the web export where to look up the assets.

# How

- Added `transformer.publicPath` config option based on an env flag
- Updated `web/deploy-script.js` to enable this env flag, and move assets to the proper folders for S3 upload

# Test Plan

See staging.


byCedric:
Tested on staging, seems to work!
image

PR #458, opened 2023-08-29
chore: improve monorepo tooling

byCedric:
# Why

With more packages being added, we should have proper tooling to support monorepos.

# How

- Added basic configuration for `start`, `lint`, `test`, and `build` focusing _only_ on packages.
- Updated workflow to optimize each package CI time

# Test Plan

See if this PR passes.


byCedric:
Superseded by #461

PR #460, opened 2023-09-06
feature(website): Update Orbit integration to fetch local server

gabrieldonadel:
# Why

Closes ENG-9990

In order to remove the need for an "Enable Orbit" option inside User Settings, we need to check if Orbit is available in the user's machine so we can automatically display the "Open with Orbit" button in the editor toolbar.

To achieve this functionality, Orbit v1 will ship with a local HTTP server that we can just fetch from the website and verify if Orbit is running. As long as Orbit is open, the "Open with Orbit" button should be displayed. 

For more context check https://github.com/expo/orbit/pull/52

# How

This PR updates the Orbit integration to fetch the localhost server using the predetermined ports (35783, 47909, 44171, 50799) and wraps all the logic of this feature in a new hook called `useOrbit`

# Test Plan

With Orbit running in the background
 

https://github.com/expo/snack/assets/11707729/4de5a1c7-bb06-4fa4-bda5-dbbbbe7ff1ab

With Orbit closed
 

https://github.com/expo/snack/assets/11707729/061b48ee-cc2a-4a5c-aff6-4f88c4a6e259

### DIFF
--- a/website/src/client/components/EditorToolbar.tsx
+++ b/website/src/client/components/EditorToolbar.tsx
@@ -1,8 +1,8 @@
 import { StyleSheet, css } from 'aphrodite';
 import * as React from 'react';
 
-import { Experiment } from '../auth/authManager';
 import { SaveStatus, SaveHistory, Viewer, SaveOptions, SDKVersion } from '../types';
+import { useOrbit } from '../utils/orbit';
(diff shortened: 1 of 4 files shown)


byCedric (approved):
lgtm

PR #464, opened 2023-09-16
feature: add monorepo tooling to minimize maintenance

byCedric:
# Why

This is another attempt to implement proper monorepo tooling to minimize maintenance as much as possible.

# How

- Added `turbo` as monorepo tool
- Configured through **turbo.json** (with **.vscode/settings.json** to set it to jsonc/json5)
- Revamped **snackager/Dockerfile** to remove hardcoded paths
  - This now uses `turbo` to remove unused workspaces, which auto-updates once they are added/removed.
  - It also reuses previously built files when possible, minimizing build time as well.
- Revamped **website/Dockerfile** to remove hardcoded paths (same as snackager)

## Docker sizes

Dockerfile | Before | After
--- | --- | ---
[snackager](./snackager/Dockerfile) | Dockerfile before this change | Dockerfile after this change
[snackpub](./snackpub/Dockerfile) | Dockerfile after this change | Dockerfile after this change
[website](./website/Dockerfile) | Dockerfile before this change | Dockerfile after this change

# Test Plan

Since this is a big change, have to fully test this on staging first.


snackager/Dockerfile
```diff
@@ -1,67 +1,67 @@
...
+FROM node:${node_version}-alpine as scoper
+WORKDIR /app
 
-RUN apk --no-cache add git openssh-client
+# Copy repository
+COPY . .
+# Prune repository and remove unused workspaces
+RUN apk --no-cache add jq && \
+  yarn global add turbo@$(jq --raw-output '.devDependencies.turbo' package.json) && \
+  turbo prune --docker --scope=snack-modules-packager
```
byCedric, line 12: This is the biggest "magic sauce" in this PR. This does a couple of things:

1. Load **package.json**
2. Get `"turbo": "<version>"` from **package.json**
3. Install this version of `turbo` globally
4. Drop _all workspaces_ that are NOT being used from the monorepo

Especially step 4 reduces the maintenance of this dockerfile. We can split up things, without having to edit this dockerfile, while maintaining smallest possible image.

snackager/package.json
```diff
@@ -37,7 +37,7 @@
     "semver-utils": "^1.1.4",
     "snack-content": "*",
     "snack-require-context": "*",
-    "snack-sdk": "*",
+    "snack-sdk": "4.1.0-rc.1",
```
Kudo, line 40: question: do you know for a reason to change this? would that help monorepo managing? if somehow we bump publish newer snack-sdk, should we change this manually? or turborepo could help that?

Kudo (approved):
basically looks good to me. good to learn the `turbo prune --docker`.
didn't see any big problem in this pr, just have to deploy and we will know.

snackager/package.json
byCedric: I guess `yarn@1` has a lot of issues connecting `"*"` versions to `-alpha|beta|rc.x`. I'll move this change from the PR to the `sdk-49` upgrade and resolve the e2e failures.

I also want to split up some workflow changes from this PR into their own.

byCedric:
@Kudo We definitely need to just try it in deployment, that's why Snack staging is already running this image 😄  And so far, its looking good.

snackager/package.json
byCedric: I had a similar issue with the website: https://github.com/expo/snack/pull/437

Kudo: that makes sense. pre-releases always act differently. that's by design of semver. i had a trick in [install-expo-modules](https://github.com/expo/expo-cli/blob/5bac95d361227497e53c99f3606294546d872040/packages/install-expo-modules/src/utils/packageInstaller.ts#L32-L37), but i don't know how to use wildcard to include a pre-release version for the case here.

maybe an alternative solution here is like publishing without pre-release (e.g. snack-sdk 4.2.0) and with `next` dist-tag.

byCedric: Yeah, that seems like a good alternative. I'd love if we can swap to some tool that allows us to be explicit in workspace versions, e.g. `"snack-sdk": "workspace:*"` (whatever is available in the workspace).

Kudo: > I'd love if we can swap to some tool that allows us to be explicit in workspace versions, e.g. "snack-sdk": "workspace:*" (whatever is available in the workspace).

that's a neat idea. hopes there are some package manager to support this semantic.

byCedric: Currently support is:

- `yarn@>=2` - [docs](https://yarnpkg.com/features/workspaces#cross-references)
- `pnpm` - [docs](https://pnpm.io/workspaces#workspace-protocol-workspace)
- `bun` - [docs](https://bun.sh/docs/install/workspaces)
- `npm` - no support, no RFC open (yet... 😄)

Kudo: good to know the workspace protocol was in real. thanks Cedric!

PR #467, opened 2023-09-17
feature(snackager): add bun as optional package manager

byCedric:
# Why

This is an experiment to see if we can speed up both Snackager and bundling in general.

# How

- Refactored install scripts to make both `yarn` and `bun` optional
- Added `bun` as optional package manager
- Added test to ensure `bun` is executed with `--ignore-scripts`

# Test Plan

TBD


byCedric:
This does not have the intended effects as I hoped, especially around e2e. Closing

PR #478, opened 2023-09-27
chore(website): update the iframe permissions when previewing on web

byCedric:
# Why

Screen wake lock was not enabled. This enables most of the APIs for users to play with in the web preview.

# How

- Found the `Permission-Policy` directives list: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy#directives
- Added all directives, while disabling a few.

# Test Plan

See staging


website/src/client/components/DevicePreview/WebFrame.tsx
```diff
@@ -41,3 +41,43 @@ const styles = StyleSheet.create({
...
+  'accelerometer',
+  'ambient-light-sensor',
+  'autoplay',
+  'battery',
+  'camera',
+  // 'display-capture',
+  // 'document-domain',
+  // 'encrypted-media',
+  // 'execution-while-not-rendered',
+  // 'execution-while-out-of-viewport',
```
Simek, line 59: Can we remove the disabled rules instead? There is a link in comment already, which leads to the full list of possible directive.

Simek (approved):
Left one suggestion, besides that LGTM, thanks for the fix! 👍

PR #491, opened 2023-10-11
[website] remove snackpub analytics

Kudo:
# Why

snackpub was fully rolled out to production and we don't need the measurement anymore. this could save the big query cost a little bit.

# How

revert partial code from #380 on website

# Test Plan

ci passed


Kudo:
going to merge this because of just minor changes.

PR #548, opened 2024-01-26
chore(website): Update Orbit deep link route

gabrieldonadel:
# Why

With the new `experienceURL` format Orbit can no longer distinguish an Expo Update from an Snack just by checking the URL, because of that we should use the `/snack` route which is specific for snacks

Closes ENG-11059

# How

Update Orbit deep link route to use `/snack`

# Test Plan

Run snack locally and launch Orbit from the button 

https://github.com/expo/snack/assets/11707729/b4f631c6-ae79-4791-8252-2185997d0e8a

### DIFF
--- a/website/src/client/utils/orbit.ts
+++ b/website/src/client/utils/orbit.ts
@@ -66,7 +66,7 @@ export function useOrbit() {
         return;
       }
 
-      const deeplink = `expo-orbit://snack/${experienceURL.replace('exp://', '')}`;
+      const deeplink = `expo-orbit://snack/?url=${encodeURIComponent(experienceURL)}`;
       if (isRunning) {
         const response = await fetchLocalOrbitServer('open', { url: deeplink });
         if (response?.ok) {
--- a/website/src/client/utils/orbit.ts
+++ b/website/src/client/utils/orbit.ts
@@ -66,7 +66,7 @@ export function useOrbit() {
         return;
       }
 
-      const deeplink = `expo-orbit://snack/?url=${encodeURIComponent(experienceURL)}`;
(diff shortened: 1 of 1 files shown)


website/src/client/utils/orbit.ts
```diff
@@ -66,14 +66,15 @@ export function useOrbit() {
         return;
       }
 
+      const deeplink = `expo-orbit://snack/${experienceURL.replace('exp://', '')}`;
```
byCedric, line 69: I think this might actually break the URL spec. What you'll get is something like this:

```
expo-orbit://snack/u.expo.dev/XXX?snack=YYY&snack-channel=ZZZ...
```

I don't think the `u.expo.dev` is allowed as path name characters.

byCedric: A good example here is Cloudflare, no idea wtf is going on here but it seems I broke their implementation of matching `CNAME` records with this type of URL https://expo-snack.app/u.expo.dev/xxx?snack=yyy

byCedric: That being said, if it works 🤷

byCedric (approved):

website/src/client/utils/orbit.ts
gabrieldonadel: Ohh I see, we should probably try to stay in the URL spec. What if we used `expo-orbit://snack/?url=https://u.expo.dev/XXX?snack=YYY&snack-channel=ZZZ...` instead?

gabrieldonadel: I've updated this PR and Orbit to support the following format `expo-orbit://snack/?url=exp://u.expo.dev/XXX?snack=YYY&snack-channel=ZZZ`

PR #558, opened 2024-02-19
refactor(website): rewrite Appetize device frame with the Appetize JS SDK

byCedric:
# Why

This reworks the full Appetize device frame, using the [Appetize JS SDK](https://docs.appetize.io/javascript-sdk/getting-started).

<details><summary>See screenshot</summary>

image

</details>

# How

- Reworked the Appetize device frame from the ground up
  - Introducing automatic device scaling
  - Per-SDK / per-appetize-instance default setting configuration, e.g. change default device for SDK 50 on the main queue but not 49 or embed
  - More native capabilities exposed but not yet implemented (such as native logs)
- Added "Device control":
  - users can reopen Snack without having to re-queue on possible crash
  - users can select a different device to run on, unleashing the full power of Appetize

# Test Plan

See staging

Tested on the React Native website, building from source with the staging environment. It looks like the screenshots below, with auto-scaling enabled. Meaning we can swap devices without having to worry about the device dimensions anymore.

android | ios
--- | ---
Android preview | iOS preview


website/src/client/utils/Appetize.ts
```diff
@@ -0,0 +1,103 @@
...
+  // ipadpro129inch5thgeneration: 'iPad Pro 12.9 (5th gen)',
+  // ipad9thgeneration: 'iPad (9th gen)',
+  nexus5: 'Nexus 5',
+  pixel4: 'Pixel 4',
+  pixel4xl: 'Pixel 4 XL',
+  pixel6: 'Pixel 6',
+  pixel6pro: 'Pixel 6 Pro',
+  pixel7: 'Pixel 7',
+  pixel7pro: 'Pixel 7 Pro',
+  // galaxytabs7: 'Galaxy Tab S7',
```
byCedric, line 52: These tablets are too big for the current panel size, this would also be an issue if we want to add a "rotate device" option. Maybe we should make the panel resizable somehow.

byCedric: See expo/snack#560

website/src/client/components/DevicePreview/DevicePreview.tsx
```diff
@@ -31,11 +31,11 @@ type Props = {
   platformOptions: PlatformOptions.PlatformOption[];
   previewRef: React.MutableRefObject<Window | null>;
   previewURL: string;
-  devices?: AppetizeDevices;
   isEmbedded?: boolean;
   sdkVersion: SDKVersion;
   sendCodeOnChangeEnabled: boolean;
   theme: ThemeName;
+  devices: AppetizeDevices[];
```
byCedric, line 38:
```suggestion
  devices: AppetizeDevices;
```

PR #559, opened 2024-02-21
fix: add codegen native component as external

byCedric:
# Why

This adds codegen native component as external, and loads it directly from the `react-native` instance in the Runtime. It's an attempt to fix `react-native-svg` on Snack.

# How

- Added external to [Snack Runtime](https://github.com/expo/snack/blob/main/runtime/src/aliases/index.tsx#L4-L31)
- Added external to [Snackager](https://github.com/expo/snack/blob/main/snackager/src/bundler/externals.ts#L8-L43)

# Test Plan

See staging, load a project with `react-native-svg` (see expo/expo#27171)


byCedric:
Seems to be working again for https://staging-snack.expo.dev/@bycedrictest/react-native-svg-example

PR #560, opened 2024-02-21
feature(website): make device preview panel resizable

byCedric:
# Why

Follow up for expo/snack#558. If we want to add device control to rotate the device, we need to allow users to resize that panel (because else it gets too small).

# How

This introduces `react-resizable-panels` and implements it for the editor and preview. Note, for embeds, this won't be enabled.

# Test Plan

See staging.

### DIFF
--- a/website/src/client/components/DevicePreview/DevicePreview.tsx
+++ b/website/src/client/components/DevicePreview/DevicePreview.tsx
@@ -151,7 +151,7 @@ class DevicePreview extends React.PureComponent<Props, State> {
     return (
       <div
         className={classnames(css(isEmbedded ? styles.embedded : styles.container), className)}
-        style={{ width }}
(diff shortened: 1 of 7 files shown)


website/src/client/components/DevicePreview/DevicePreview.tsx
```diff
@@ -218,9 +218,10 @@ export default withThemeName(DevicePreview);
...
     position: 'relative',
-    maxWidth: '50%',
-    overflowX: 'hidden',
-    overflowY: 'auto',
+    width: '100%',
+    // maxWidth: '50%',
+    // overflowX: 'hidden',
+    // overflowY: 'auto',
     display: 'none',
     flexDirection: 'column',
```
byCedric, line 226:
```suggestion
    width: '100%',
    display: 'none',
    flexDirection: 'column',
```

PR #570, opened 2024-03-06
chore(ci): Google Cloud Workload Identity Federation

nicknovitski:
# Why

I hope this will let us remove the service account key from our GitHub Actions secrets.

# Test Plan

I was able to succesfully deploy snackager to staging, which suggests the other deployments will work too. https://github.com/expo/snack/actions/runs/8271897475/job/22632693926?pr=570

### DIFF
--- a/.github/actions/setup-google-cloud/action.yml
+++ b/.github/actions/setup-google-cloud/action.yml
@@ -2,53 +2,33 @@ name: Setup Google Cloud SDK
 description: Prepare Google Cloud SDK in GitHub Actions
 
 inputs:
-  project-id:
-    description: Google Cloud SDK project id
-    default: exponentjs
-
-  project-zone:
-    description: Google Cloud SDK compute zone
-    default: us-central1
-
-  project-cluster:
-    description: Google Cloud SDK container cluster
-    default: general-central
-
-  service-key:
-    description: Google Cloud SDK service_account_key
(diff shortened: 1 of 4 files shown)


.github/actions/setup-google-cloud/action.yml
```diff
@@ -2,53 +2,33 @@ name: Setup Google Cloud SDK
...
-    default: general-central
-
-  service-key:
-    description: Google Cloud SDK service_account_key
-    default: ''
-
-  export-default-credentials:
-    type: boolean
+  auth:
     description: If it should export the default credentials
```
ide, line 6:
```suggestion
    description: If it should authenticate with Google Cloud and the Kubernetes cluster
```
Also, shouldn't this type be a boolean and look for true/false instead of empty strings?

ide (changes requested):
Requesting some small changes to the parameter type

.github/actions/setup-google-cloud/action.yml
nicknovitski: That would be nice, but [action inputs do not have types](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs), they are all strings.  You can type the inputs to re-usable workflows, but not actions.

ide (approved):

PR #579, opened 2024-05-07
refactor(snack-content): set SDK 51 as default

byCedric:
# Why

This should be the default when the blog post is released.

# How

- Configured SDK 51 as default through `snack-content`.

# Test Plan

See staging


nitinmuk:
@byCedric I am trying to use latest `snack-sdk` version 5.0.0 which uses `snack-content` 2.0.0 and it doesnot support SDK 51. So, I am wondering if you plan to release new versions with SDK 51 support soon ?

PR #601, opened 2024-08-15
feature(website): Allow the initially selected file to be specified

### LINKED ISSUE #326: Allow embedded Snacks to view code other than the `App.js` file
### Do you understand that any feature requests opened in the core Expo Snack repository will be closed?

Yes

robwalkerco:
Fixes #326 

# Why

When embedding a snack, the embedded view only shows the `App.{js,tsx)` file in the embedded view. To view other files, the user needs to open the snack on snack.expo.dev so that they can view the file browser.

This limitation can cause issues when it's desired to show the contents of a file other than `App.{js,tsx)` in an embedded snack, such as when a snack is using `@shopify/react-native-skia` and needs to be able to render using `react-native-web`. In this scenario, the embedded snack ends up showing just the code needed to setup Skia for the web, rather than the code used to render the Skia component.

With the ability to set the `initiallySelectedFile` param, the embedded snack can be set to show the most relevant file.
When opening the full snack view using the icon, the param will be passed through and used to select the initial file when the file browser is visible.

If an invalid filename is specified, it will be ignored.

# How

I was able to add this feature by running the website locally (I couldn't get the `http://snack.expo.test` setup to work, so I just tested by running the website itself locally.

After investigating how other params like `sourceUrl` work, I was able to add the `initiallySelectedFile` param in a similar fashion.

# Test Plan

I was able to test while running the website locally.
As I was running against `localhost:3011`, I had to manually adjust the [`isEmbedded`](https://github.com/expo/snack/blob/a60fe546ae76e58c13538ba6ba552fa624b0654d/website/src/client/components/Router.tsx#L35) logic so that I could test on the embedded and not-embedded clients.

Recording of using this param on the embedded client

https://github.com/user-attachments/assets/d9510c51-a5cc-4871-8180-85931230e122

Recording of using this param on the full client

https://github.com/user-attachments/assets/f0d65a25-ea3f-414c-a388-b3008061af69

Note: As I needed to adjust the code to the embedded vs not embedded client, testing linking from the embedded client to the full client was not fully possible, but by setting `hideQueryParams=true` I could check that the new `initiallySelectedFile` was being added to the URL when using the button that inks to the full client.

### DIFF
--- a/docs/embedding-snacks.md
+++ b/docs/embedding-snacks.md
@@ -44,6 +44,7 @@ The `embed.js` script scans the DOM and populates any elements containing a `dat
 | `data-snack-preview`| Shows or hides the preview pane. Defaults to `true` using `embed.js` Snacks. Valid values: `true`, `false`. |
 | `data-snack-sdkversion` |  The Expo SDK version to use (eg. `38.0.0`). Defaults to the latest released Expo SDK version. |
 | `data-snack-supportedplatforms` |  The platforms available for previewing the Snack. Defaults to `mydevice,ios,android,web` when not specified. |
+| `data-snack-initially-selected-file` | The filename of the file to show in the embedded snack. |
 | `data-snack-theme` | The theme to use, `light` or `dark`. When omitted uses the theme that was configured by the user (defaults to `light`). |
 | `data-snack-device-frame` | If the outline of the device should be rendered or not. Valid values: `true` or `false`. |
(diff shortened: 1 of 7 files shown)


zibs:
Hey @byCedric -- Just wanted to follow up here and get your thoughts on this one!

zibs:
Hey @byCedric let us know if we can do anything to help this along!

PR #621, opened 2025-01-03
fix(snackager): add e2e bundling test for `react-native-webview@13.12.5`

byCedric:
# Why

`react-native-webview@13.12.5` imports the vendored event emitter file from `react-native`. This file unfortunately uses syntax that can't be parsed by the Babel flow parser (`[key in keyof TEventToArgsMap]`). To combat that, this PR adds the Hermes babel plugin for flow code.

# How

- Added `babel-plugin-syntax-hermes-parser@0.26.0` to Snackager

# Test Plan

See E2E tests


Kudo:
thanks for fixing the problem. the EventEmitter from react-native-webview was imported by me 😅

byCedric:
So... IT WAS YOU!

Just kidding, Snackager should be able to parse any type of Flow code anyways -- happy to fix it.

PR #675, opened 2026-02-25
Upgrade snackager TypeScript to 6.0 and target Node 22

ide:
## Why

Snackager runs on Node 22.22 but the TypeScript configuration still extended `@tsconfig/node16` with `es2021` lib and target. This meant newer ES features like `ErrorOptions` (es2022) and other post-2021 type definitions were unavailable.

## How

Upgrade `typescript` from `^5.3.3` to `^6.0.3` and switch the tsconfig base from `@tsconfig/node16` to `@tsconfig/node22`.
- Set `module` and `moduleResolution` to `nodenext` to match the modern preset, and drop the deprecated `baseUrl` since no `paths` mapping anchors it and no source files rely on it for resolution.
- Add an explicit `types: ["node", "jest"]` because TypeScript 6 only auto-includes `@types/*` packages from the project's immediate `node_modules/@types/`, and `@types/jest` — which contributes only ambient globals — is hoisted by Yarn workspaces to the repo root rather than installed locally.

In `snackager/src/cli.ts`, the stricter `Uint8Array` types in es2024 surfaced a `Buffer` incompatibility with `fs.writeFileSync`, fixed by casting through `unknown` to `Uint8Array` (the type declarations are being pedantic because of SharedArrayBuffer).

In `snackager/package.json`, prefix the `tsc` invocations with `node_modules/.bin/` and add a `tsc` script so that `yarn tsc` runs the correct, local version of tsc.

## Test Plan

`yarn tsc` passes.

### DIFF
--- a/snackager/src/cli.ts
+++ b/snackager/src/cli.ts
@@ -77,7 +77,7 @@ async function main(): Promise<void> {
       if (!fs.existsSync(path.dirname(filePath))) {
         fs.mkdirSync(path.dirname(filePath));
       }
-      fs.writeFileSync(filePath, buffer);
+      fs.writeFileSync(filePath, buffer as unknown as Uint8Array);
       const info = getBundleInfo(filePath, buffer);
       logger.info({ file: filePath, ...info }, `wrote file to ${path.basename(filePath)}`);
     });
--- a/snackager/package.json
+++ b/snackager/package.json
@@ -6,12 +6,13 @@
   "scripts": {
     "start": "REDIS_ENVIRONMENT=staging scripts/dev.sh",
     "start:production": "REDIS_ENVIRONMENT=production scripts/dev.sh",
-    "build": "tsc --project tsconfig.build.json",
+    "build": "node_modules/.bin/tsc --project tsconfig.build.json",
+    "tsc": "node_modules/.bin/tsc",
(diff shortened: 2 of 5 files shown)


byCedric (approved):

snackager/src/cli.ts
```diff
@@ -77,7 +77,7 @@ async function main(): Promise<void> {
       if (!fs.existsSync(path.dirname(filePath))) {
         fs.mkdirSync(path.dirname(filePath));
       }
-      fs.writeFileSync(filePath, buffer);
+      fs.writeFileSync(filePath, new Uint8Array(buffer));
```
byCedric: Is this `Uint8Array` wrapping necessary? From what I can see, the buffer is already somewhat in a special memory, and [`fs.writeFileSync` should be able to write from Buffer directly](https://nodejs.org/api/fs.html#fswritefilesyncfile-data-options)

ide: The type declarations (either in `@types/node` or the types that come with tsc) are a bit pedantic. It has to do with SharedArrayBuffers possibly causing conflicts and there are several GH issues complaining about this.

I decided to use `as unknown as Uint8Array` to avoid changing runtime behavior. Plus `new Uint8Array(buffer)` does a copy I believe - there is a zero-copy way to do this but you have to specify the offset and length and I think we shouldn't put the type system ahead of the runtime.

PR #684, opened 2026-05-08
[snack-sdk][website][snackager] Migrate API server hostname

byronkarlen:
# Why

For cookie-based auth to work, the API server url needs to be updated from exp.host to api.expo.dev (which shares the top-level domain with snack.expo.dev). api.expo.dev accepts requests in a different format than exp.host.

# How

snack-sdk

- Changed default apiUrl to api.expo.dev
- Use the modern path format
- Updated docs since this is a breaking change

snack-proxies

- removed www from snack-proxies. No longer used. Instead, snack website runs against proxy-staging-api.expo.test, but this can be overriden with API_SERVER_URL

snackager

- passes new api server url to snack-sdk

website

- passes new api server url to snack-sdk
- updates its own API requests

# Test Plan
Used snack.expo.test locally and verified that thing worked as expected, but we don't have e2e test coverage for this.

PR #681, opened 2026-05-08
[website] Fix test failing locally

byronkarlen:
# Why

FileListEntry.test.tsx fails locally but passes on Linux CI. The test indirectly captures KeyMap.Meta's integer value, which KeybindingsManager.tsx derives from navigator.platform at module load. Jest's node test environment historically had no globalThis.navigator, so the snapshot was recorded with the non-Mac value. Node 21+ now exposes a built-in globalThis.navigator whose platform reflects the host OS, so on Macs the the snapshot mismatches.

# How

Stub globalThis.navigator injest/unit.setup.js with a fixed non-Mac value so snapshot stays deterministic across hosts.

# Test Plan

Tests pass.

PR #685, opened 2026-05-08
[snack-sdk][website] Move to cookie-based auth

byronkarlen:
# Why

Remove client-side reads of the session cookie. Rely on `credentials: 'include'` instead.

# How

snack-sdk

- added a useCookieAuth option that defaults to false for existing consumers

website

- sets snack-sdk's useCookieAuth to true
- refactored authManager to get rid of legacy auth code, and use credentials: 'include' with direct fetches to website

To make these changes, I had claude audit all of website's requests (either direct or through snack-sdk) to the API server. Then, I had it determine which of these endpoints needed potentially authenticated contexts and added support for `credentials: 'include'`.

# Test Plan

Tested out auth flow locally against local instance of website and staging api. Not super familiar with snack so I'm not confident my QA was comprehensive.


byronkarlen:
* **#681** <a href="https://app.graphite.com/github/pr/expo/snack/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#685** <a href="https://app.graphite.com/github/pr/expo/snack/685?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/snack/685?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#684** <a href="https://app.graphite.com/github/pr/expo/snack/684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/snack-sdk/src/DevSession.ts
```diff
@@ -29,16 +32,20 @@ export default class DevSession {
...
 
     // Close
-    const isCloseUser =
-      prevState.user &&
-      (!state.online || state.url !== prevState.url || state.user !== prevState.user);
+    // Under cookie auth the SDK does not hold a user object — the server
+    // identifies the user from the cookie — so use online-state as the
+    // "had a user-keyed session" signal in that mode.
+    const isCloseUser = this.useCookieAuth
+      ? prevState.online && (!state.online || state.url !== prevState.url)
```
ide: I'm not sure this is right - would `state.online` be set in the anonymous scenario? Looking at the `this.close()` call below, we don't send the user anyway when using cookies.

I think the current code works but it ends up being harder to read because of the history of how we ended up here.

Logically it's like we want:

```
if (useCookieAuth) {
  // cookie-specific logic here...
  this.close(...)
} else {
  // old logic here...
  this.close(...)
}
```

ide (changes requested):

packages/snack-sdk/src/DevSession.ts
byronkarlen: Yes `state.online` could be set in the anonymous scenario. Agreed confusing. I just changed it to be clearer. The logic is the same though.

ide (approved):

PR #694, opened 2026-07-02
[production] Pin snack / snackager / snackpub to the prod node pool

hungtuanz:
## What
Move the **production snack, snackager, snackpub** production workload(s) off the shared NAP pools onto the dedicated **`prod`** GKE node pool (general-central): `nodeSelector: pool=prod` + toleration `dedicated=prod:NoSchedule`.

## Why
Node-pool consolidation — production workloads belong on the `prod` pool, off NAP.

## Compatibility (verified against the live cluster)
- **Stateless** — no PVCs, so no pd-balanced/Hyperdisk attach issue on the c4d-standard-8 `prod` nodes.
- `prod` pool taint = `dedicated=prod:NoSchedule` → the added toleration is sufficient.
- Additive patch (base has no nodeSelector) → pool-only selector, no machine-family/iam pins. Production overlay render validated; staging overlays unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/snackager/k8s/production/deployment-prod-pool.yaml
+++ b/snackager/k8s/production/deployment-prod-pool.yaml
@@ -1,0 +1,13 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: snackager
+spec:
+  template:
+    spec:
+      nodeSelector:
+        pool: prod
+      tolerations:
+      - key: dedicated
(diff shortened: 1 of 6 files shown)


hungtuanz:
@byCedric could I get a review please?

